PR #9, opened 2015-01-21
Fixes ETag quoting and strips white space from generated Etag.

lawlesst:
The Etags being generated aren't quoted.  The initial quote was missing.   

`Language-specific ETAG = 3a6d8837c0078966en_US en "`

This change wraps the ETag in quote as [recommended/specified](http://tools.ietf.org/html/rfc7232#section-2.3).  This also removes trailing white space that was being added when appending the locales to rawEtag.  Although it doesn't seem to be invalid, I'm also striping whitespace from locales for consistency.

### DIFF
--- a/webapp/src/edu/cornell/mannlib/vitro/webapp/filters/CachingResponseFilter.java
+++ b/webapp/src/edu/cornell/mannlib/vitro/webapp/filters/CachingResponseFilter.java
@@ -282,13 +282,13 @@ public class CachingResponseFilter implements Filter {
 		@SuppressWarnings("unchecked")
 		List<Locale> locales = EnumerationUtils.toList(req.getLocales());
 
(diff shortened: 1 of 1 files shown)


j2blake:
Thanks for catching this, Ted.

Either of these will append a quote to a StringBuilder:
`builder.append('"');`
`builder.append("\"");`
That's why the trailing quote is successfully appearing.

However, while this will create a StringBuilder that holds a quote character
`new StringBuilder("\"");`
This will create an empty StringBuilder with an initial capacity of 34 characters
`new StringBuilder('"');`
[http://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html#StringBuilder%28int%29]

j2blake:
I'll merge this into the develop branch.

PR #10, opened 2015-02-10
Problem with paused indexer

grahamtriggs:
The SearchIndexer is starting PAUSED. Requesting a reindex just puts it into the deferred queue, but doesn't start the indexer.

This patch will unpause the index when you request a rebuild, avoiding the problem of the index request not being processed.

It would probably be better to ensure that the indexer starts UNPAUSED, and that the reindex screen provides feedback as to when and why the indexer is paused, and when it will be unpaused.

### DIFF
--- a/webapp/src/edu/cornell/mannlib/vitro/webapp/search/controller/IndexController.java
+++ b/webapp/src/edu/cornell/mannlib/vitro/webapp/search/controller/IndexController.java
@@ -172,6 +172,7 @@ public class IndexController extends FreemarkerHttpServlet {
 
 	private void requestRebuild() {
 		indexer.rebuildIndex();
+		indexer.unpause();
 	}
 
 	private Map<String, Object> buildStatusMap(SearchIndexerStatus status) {


j2blake:
Graham -- this is really cool. I'm delighted that someone is checking my work on the indexer.
I had already had my nose rubbed in the fact that starting in PAUSED mode is not a good idea, even though SearchIndexerSetup was calling unpause() immediately after calling startup(). In commit 9f9cea06e360aa9c (Feb 7), I restructured this code to use two flags: paused and started, rather than attempt to make one flag suffice.

Before or after that commit, however, I would expect that the SearchIndexer would be UNPAUSED by the time IndexController was invoked. The only exception I know of would be if ABoxRecomputer was in the process of reinferencing. In that case, a premature call to unpause() would be counter-productive.

My impulse is to take your last suggestion, adding a message parameter to the pause() method, which would then be reported as part of the status. I might also be inclined to add a "UNPAUSE" button to the IndexController page, rather than incorporate it as a silent side effect of the "REBUILD" button.

Your thoughts?

Jim

grahamtriggs:
Not exactly checking, just trying to figure out why it didn't do anything when I kept clicking the button, and then occassionally kicking off all of those deferred requests for no apparent reason! :)

I think the real solution here is to have a unified task executor, so that recompute and search tasks can be submitted to a single queue and then indexes can't run whilst reinferencing, and vice versa.

Then the admin pages that provide buttons for kicking off a rebuild / recompute would poke the queue and not provide the options to add a new task of that type if there is already one queued (or being processed).

That would get rid of the deferred queue and the possibility of lining up hundreds of seach reindexes from the page not giving any feedback for the indexer being paused (which can be quite annoying).

Ultimately, it would probably be good to get rid of a lot of the direct executor code, and go through Quartz instead - which would then lead to the possiblity of being able to set up these jobs to run on a timed schedule.

j2blake:
Graham;

Thanks for the tip on Quartz. I will have a look at it. Your suggestion about a unified task executor is appealing; I am copying this note to Brian Lowe, who has handled most of our inferencing code. A unified task executor will not happen now, since we are trying to complete release 1.8. However, perhaps we can coordinate this in the near future.

Jim

arademaker:
@grahamtriggs 1+ vote to your idea. The unification of task executor would make the application more robust and scalable IMHO.

j2blake:
The proximate cause of this problem was an oversight in https://github.com/vivo-project/Vitro/commit/9f9cea06e360aa9c586d28c7858d7e5eac664ef5, corrected in 34b01ee1988152006f1294cd05859498ac3cf20c, which removes the immediate need for this pull request.

A JIRA issue https://jira.duraspace.org/browse/VIVO-967 has been created to track the topics raised in this discussion.

grahamtriggs:
Jim, has 34b01ee been tested? I've just updated my codebase, and I'm still having the same problem with the Indexer starting paused and not executing rebuilds.

j2blake:
Yes, it has been tested. Is it possible that inferencing is in process? A sequence of messages such as this:

2015-02-11 10:54:28,889 INFO  [ABoxRecomputer] Recomputing ABox inferences.

2015-02-11 10:54:29,676 INFO  [IndexHistory] PAUSE, 2/11/15 10:53 AM, []

would indicate that the indexer is legitimately paused.

grahamtriggs:
Hi Jim,

You're right - there was a recompute happening. It was a completely fresh install.

However, it appears the recompute had finished by the time the indexer had started, but it was left in a PAUSED state, so the "Rebuild index" was exhibiting the same problem as before.

I think before this is released, it at least needs to only queue up one rebuild in the deferred list (repeatedly hitting the button and queueing up rebuilds "blind" is a problem).

j2blake:
Closed. Superseded by #14

PR #11, opened 2015-02-10
removed extra } from title attributes in vitro templates

nateprewitt:
This is just a minor syntax fix. The hover title was appending an extra } to the end of the phrase "View the profile page for ____" on all group pages. I'm also including a related fix to affected templates in VIVO in a separate pull request.


j2blake:
Thanks, Nate.

PR #12, opened 2015-02-11
Unpause indexer before rebuild, change listener startup order

grahamtriggs:
If you start up an empty VIVO (/Vitro) application, then the ABoxRecompute will start, pausing the index, then restarting it after.

However, before the restart, it was attempting to schedule an index rebuild, which threw a NullPointerException, as the SearchIndexer had not started, and a DAO was null.

I've put the unpause before the rebuildIndex, so that any exceptions thrown by the rebuildIndex won't cause the search indexer to be stuck in a paused state.

Also, moved the reasonor listener to after the SearchIndexer, so that the searchindexer is started first, ensuring that it is initialized before the reasoner, so that the rebuildIndex won't throw a NullPointerException.


j2blake:
Thanks, Graham - it will likely be a couple of days before I can look at these changes

PR #14, opened 2015-02-15
Allow searchindex tasks to be created before startup, simplify indexlistener

grahamtriggs:
This pull request encompasses a number of related changes:

1) Push implementation of SearchIndex.Tasks into inner classes, and introduce an IndexerConfig adapter that allows accessing of the lists, dao, etc. to be deferred until the Task is about to run.

- this avoids a NullPointerException that can occur when creating a Task before the SearchIndexer has started.

2) Add a method to allow pausing without deferring tasks / scheduling a full rebuild when unpaused. Useful for tasks like ABoxRecomputer which is going to schedule a full rebuild anyway, to prevent accumulations of unnecessary tasks / changes when they are never going to be processed.

(Observed reduction of recompute memory usage by about 0.5GB - 1GB on a large dataset)

3) Simplify IndexChangeListener, avoiding having two places buffering changes when the indexer is paused. (On unpausing, if they survive, a few update tasks with smaller changesets should require less memory to process)

4) Implement RiotReader and Model.asStatement to reduce number of Models created, and remove need to disconnect Statements from the Model they were created as part of.

NB: Not sure why I haven't seen this before, but on develop I've stated getting log messages like:

2015-02-15 15:15:43,748 ERROR [riot] [line: 7, col: 95] Unrecognized: [DOT]
2015-02-15 15:16:10,614 ERROR [riot] [line: 1, col: 113] Triples not terminated by DOT
2015-02-15 15:16:10,615 ERROR [riot] [line: 2, col: 3 ] Unrecognized: SUBMITTED

when adding / editing nodes - particularly where there are empty assertions (e.g. a full name without middle name). This is coming from the parseN3ToRDF() method.

Log4J settings look to suppress these with:

log4j.logger.org.openjena.riot=FATAL

but now it should be:

log4j.logger.org.apache.jena.riot=FATAL

for Jena 2.10.0+

### DIFF
--- a/webapp/src/edu/cornell/mannlib/vitro/webapp/modules/searchIndexer/SearchIndexer.java
+++ b/webapp/src/edu/cornell/mannlib/vitro/webapp/modules/searchIndexer/SearchIndexer.java
@@ -40,10 +40,18 @@ public interface SearchIndexer extends Application.Module {
 	 * unpause(). Fires a PAUSED event to listeners.
 	 * 
 	 * This call has no effect if already paused, or if called after shutdown.
-	 */
+     */
 	void pause();
 
-	/**
+    /**
+     * Stop processing new tasks. Requests will be ignored and the index rebuilt when unpaused.
+     * Fires a PAUSED event to listeners.
+     *
+     * This call has no effect if already paused, or if called after shutdown.
+     */
+    void pauseWithoutDeferring();
+
+    /**
 	 * Resume processing new tasks. Any requests that were received since the
 	 * call to pause() will now be scheduled for processing. Fires an UNPAUSED
 	 * event to listeners.
--- a/webapp/src/edu/cornell/mannlib/vitro/webapp/reasoner/ABoxRecomputer.java
+++ b/webapp/src/edu/cornell/mannlib/vitro/webapp/reasoner/ABoxRecomputer.java
@@ -104,7 +104,7 @@ public class ABoxRecomputer {
         }
         try {
             if  (searchIndexer != null) {
-                searchIndexer.pause();
+                searchIndexer.pauseWithoutDeferring();
             }
             recomputeABox();
         } finally {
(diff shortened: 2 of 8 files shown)


j2blake:
graham -- I'll have a look.

j2blake:
There is a lot here, and a lot to like, but also some issues. I am going to merge this, but I will add some changes by means of a separate commit.

Changes to SearchIndexerImpl and related task classes:
- This is a nice handling of a nasty situation. I wish that all of this chicanery were not required. Perhaps in the future we can dispense with it. (https://jira.duraspace.org/browse/VIVO-683) Now, however, it is necessary, and this appears to be a rigorous solution. My own approach was flawed in multiple respects. Thank you.

Changes to SearchIndexer:
- I am renaming pauseWithoutDeferring() to pauseInAnticipationOfRebuild(), which I believe is more to the point. I am also changing the comments to more accurately reflect the effect of the method.

Changes to IndexingChangeListener:
- Good: I appreciate the serialization improvements in IndexingChangeListener. I am not as familiar with the RIOT packages as I should be. 
- Not good: the "pause" logic cannot be removed. Your commit message says that several small batches of changes may be more efficient than one large batch, but that is no true. When calls to pause() and unpause() are made without a call to rebuildIndex(), it is usually because a file of related triples is being added or deleted. Breaking that indexing task into small batches leads to repetitive re-indexing of the same record. This is the precise reason why IndexingChangeListener creates batches.

grahamtriggs:
Jim - thanks for the review and comments. I should clarify.

When I say that small batches are more efficient, I mean memory efficient. Making one very large collection of Statements, which then becomes a very large selection of URIs before any other work takes place is going to run out of memory eventually. Breaking it down into smaller batches will allow larger workloads to complete, even if it takes a bit longer.

Getting rid of the pause logic is only with respect to the handling of a pause request to the indexer. This is not the same as the ticker in the listener, that batches up requests that occur in a short space of time.

I've left that ticker in place to create small batches of changes to pass to the indexer, but I removed the bit where it listens to the indexer being paused.

I think that if we want the indexer to combine these tasks whilst it is paused, then it would be better to combine those tasks in the indexer, rather than the listener. But it should have a limit to the size of the tasks it creates.

j2blake:
Agreed. Further, if a batch of changes becomes large enough, the indexer could decide that a full rebuild would likely be faster. But that is for the future. Unless you see problems, I'm going to say that this is good enough for this release.

PR #15, opened 2015-02-16
Update log4j.properties

grahamtriggs:
Add org.apache.jena.riot to log4j, as the packages moved in 2.10.x (suppress riot errors in ProcessRDdfForm.parseN3ToRDF which were previously suppressed by log4j.logger.org.openjena.riot=FATAL)


j2blake:
graham - why would we not remove the restriction on org.openjena.riot? Isn't it obsolete?

grahamtriggs:
In 2.10.1, there are still classes under org.openjena.riot. The Jena Model code has been modified to use org.apache.jena.riot (which is where we are triggering the errors, and need to include it in log4j.properties).

org.openjena.riot has been removed from the current codebase (its presence in 2.10.1 must be transitional).

It looks like we can remove org.openjena.riot from the log4j.properties, but I don't see that it causes harm to retain it whilst we are using a transitional release of Jena.

j2blake:
Thanks for clarifying.

PR #16, opened 2015-02-17
Rework pause control for SearchIndexer

grahamtriggs:
Opening pull request for discussion more than being "complete" (although it is functional in current state).

Thinking about the pause / unpause interactions on SearchIndexer, and there are two distinct problems:

1) The inner class is synchronized, but the outer isn't - so all of the logic inside the pause() / unpause() methods could potentially clash

2) There are two areas currently that pause the indexer - the reasoner and the sparql update endpoint. However, there is only a boolean state, so the first thing to hit "unpause" will free up the searchindexer. That could be the Sparql endpoint, whilst a reinference is still a long way from completion.

In this patch, I've pulled the pause logic out of the inner class, on to the main api, and have synchronized it at that level.

Further, the paused indicator is a count, not a flag. Every entry to pause increments the flag, every unpause decrements it. PAUSE event is fired when it first transitions into paused, and UNPAUSE when the count returns to zero. Overlapping Sparql / reinference tasks won't cause the indexer to unpause prematurely.

When paused, scheduleStatements / scheduleUris now accumulate the passed Statements / uris in collections, rather than immediately scheduling a search task. The task is created when the scheduler unpauses.

A couple of additional notes:

1) This puts the code within striking distance of doing max batch sizes for statement /.uris (which will ease memory pressures).

2) For pauseInAnticipationOfRebuild() to work as intended, then the paused flag/handling needs to be removed from the listener. If not, then everything will be queued up in the listener without ever reaching the indexer, so it won't get the opportunity to know that a full rebuild will occur (or to make decisions about batch sizes in the future).

With the changes in the indexer included here, then the updates will be queued for a single task anyway when paused, unless/until a rebuild is scheduled.

The Listnener retains the ticker, so that frequent notifications of changes when the indexer is unpaused still gets batched up into a single index task.

### DIFF
--- a/webapp/src/edu/cornell/mannlib/vitro/webapp/modules/searchIndexer/SearchIndexer.java
+++ b/webapp/src/edu/cornell/mannlib/vitro/webapp/modules/searchIndexer/SearchIndexer.java
@@ -40,18 +40,19 @@ public interface SearchIndexer extends Application.Module {
 	 * unpause(). Fires a PAUSED event to listeners.
 	 * 
 	 * This call has no effect if already paused, or if called after shutdown.
-     */
+	 */
 	void pause();
 
-    /**
-     * Stop processing new tasks. Requests will be ignored and the index rebuilt when unpaused.
-     * Fires a PAUSED event to listeners.
-     *
-     * This call has no effect if already paused, or if called after shutdown.
-     */
-    void pauseWithoutDeferring();
+	/**
+	 * Stop processing new tasks. If any request is received while the indexer
+	 * is paused, the request will be ignored, but the index will be rebuilt
+	 * when unpaused. Fires a PAUSED event to listeners.
+	 * 
+	 * This call has no effect if already paused, or if called after shutdown.
+	 */
+	void pauseInAnticipationOfRebuild();
 
-    /**
+	/**
 	 * Resume processing new tasks. Any requests that were received since the
 	 * call to pause() will now be scheduled for processing. Fires an UNPAUSED
 	 * event to listeners.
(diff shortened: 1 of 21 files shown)


j2blake:
Graham;

The short answer is that this is not a high priority for release 1.8. You make valid points, but they represent edge cases that are easily avoided. After the release, it may make sense to look at these issues in a larger context.

That larger context is the overall design of the indexer. I do hope that the indexer code is cleaner than it was in release 1.7, but there are still some very basic questions to be answered.

The indexer logic would be much simpler if we could remove the need to accept tasks before the indexer is started. Would it be possible to re-architect the VIVO startup sequence to insure that the reasoner becomes idle before the search indexer is started? In that case, could we dispense with pre-startup tasks except for a single flag called “rebuildAtStartup”? 

We have ignored the possibility that the search engine is not yet available when the indexer starts, even though that is frequently the case. Again, the indexer code would be made much more complex by trying to deal with this possibility.  How can we insure that this does not happen?

There may be other ways to keep the indexer logic clean. I doubt that such solutions will be found by dealing with the code a line at a time, or a method at a time. And the proliferation of flags leads to code that is error-prone.

Some initial thinking has already been done on this: https://jira.duraspace.org/browse/VIVO-683

We prefer to use JIRA issues for conversations, rather than pull requests. It might also be more effective for us to arrange a video call, after the release.

grahamtriggs:
Thanks Jim,

Good point about the search engine - however, it is more than saying "frequently". If you are hosting Solr and VIVO within the same application server, you may be guaranteed that solr will not receive updates until the entire server startup (including deployed applications) is complete.

I'll add it to the ticket, but the "simplest" approach is to not worry about what is / isn't available during startup, but address what should happen if an external dependency is unavailable. You can't ever really guarantee that Solr is available during the lifetime of the application / servlet container, and building queueing / retrying mechanisms to cope with that scenario at any time would also cover the startup issues.

j2blake:
Graham -- let's wait on this until it is mature. I don't think that the "pull request" mechanism is the correct place for discussions. Of course, the soon-to-be tech lead may disagree ;-)

PR #17, opened 2015-02-19
Allow preliminary "content negotiation" in VIVO SPARQL endpoint; VIVO-977

nateprewitt:
The conditional clauses contents could be moved to a more robust method to handle multiple Accept parameter values with weighting but this works for basic header modification.

Tested for empty parameter and malformed header, empty parameter and missing header, and non-existent mime-types.


nateprewitt:
Closing request to remove merge garbage, will resubmit. Sorry.

PR #18, opened 2015-02-19
Allow preliminary "content negotiation" in VIVO SPARQL endpoint; VIVO-977 

nateprewitt:
The conditional clause's contents could be moved to a more robust method to handle multiple Accept parameter values with weighting but this works for basic header modification.

Tested for empty parameter and malformed header, empty parameter and missing header, and non-existent mime-types.

See [VIVO-977](https://jira.duraspace.org/browse/VIVO-977) on JIRA for issue information.

### DIFF
--- a/webapp/src/edu/cornell/mannlib/vitro/webapp/controller/admin/SparqlQueryController.java
+++ b/webapp/src/edu/cornell/mannlib/vitro/webapp/controller/admin/SparqlQueryController.java
@@ -11,6 +11,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Enumeration;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


j2blake:
Nate -- oops, I saw your JIRA issue but not the pull request. I'll have a look. -- Jim

j2blake:
Nate, the SparqlQueryController is intended to respond to the form input from admin-sparqlQueryForm.ftl. This form does not provide Accept headers, so the controller does not consider them.

In earlier releases, there was an attempt to make SparqlQueryController more of a general-purpose access point. In release 1.7, the SPARQL Query API was introduced, implemented by SparqlQueryApiController, and SparqlQueryController was pared back to suit its original purpose.

SparqlQueryApiController attempts to handle Accept headers properly, including best-fit analysis using ContentTypeUtil.bestContentType(). As such, I'm inclined to close this pull request without merging.

The issues that you raise in VIVO-977 are still valid, and should be addressed, but that is a separate question.

nateprewitt:
Hey Jim,

ContentTypeUtil.bestContentType() definitely looks like something close to what I was thinking for proper with proper type/q param weighting. 

As for the SparqlQueryController, Accept Headers are provided regardless since it's an HTTP request, but the endpoint chooses to ignore them for the adhoc GET param substitution. This creates some of the same issues I brought up in VIVO-980. That may be fine though, if SparqlQueryController isn't considered the proper access point, then I'll move my testing over to the Query API instead.

Thanks for taking a look at this Jim

PR #21, opened 2015-07-01
Fix string conversion

patrickmcelwee:
I ran into this problem when running VIVO on MarkLogic, which always returns
type annotations for strings (which is consistent with the SPARQL spec, but
different from the Jena SDB behavior)

Strings of the form, "1^^http://www.w3.org/2001/XMLSchema#string", were
being created and then not correctly parsed into ints. (The
RDFNode.toString() method maintains the type annotation.)

To fix, make sure we remove the type information if it exists before trying to
parseInt.

### DIFF
--- a/webapp/src/edu/cornell/mannlib/vitro/webapp/controller/individual/IndividualResponseBuilder.java
+++ b/webapp/src/edu/cornell/mannlib/vitro/webapp/controller/individual/IndividualResponseBuilder.java
@@ -290,7 +290,7 @@ class IndividualResponseBuilder {
             ResultSet results = QueryUtils.getLanguageNeutralQueryResults(queryStr, vreq);
             if (results.hasNext()) {
                 QuerySolution soln = results.nextSolution();
-                String countStr = soln.get("labelCount").toString();
+                String countStr = soln.get("labelCount").toString().replaceFirst("\\^.*", "");
                 theCount = Integer.parseInt(countStr);
             }
         } catch (Exception e) {
@@ -309,7 +309,7 @@ class IndividualResponseBuilder {
                ResultSet results = QueryUtils.getLanguageNeutralQueryResults(queryStr, vreq);
                if (results.hasNext()) {
                    QuerySolution soln = results.nextSolution();
-                   String countStr = soln.get("languageCount").toString();
+                   String countStr = soln.get("languageCount").toString().replaceFirst("\\^.*", "");
                    theCount = Integer.parseInt(countStr);
                }
            } catch (Exception e) {


j2blake:
Hi, Patrick -- sorry it's taken me so long to respond. This slipped past me while I was on vacation.

Are you saying that MarkLogic doesn't distinguish between plain (untyped) literals and String literals?

patrickmcelwee:
Jim,

Now you caught me on vacation. Sorry for my delay.

MarkLogic does distinguish between typed and untyped literals. But what I believe is happening (though I have not actually verified this in a robust way) is that when Jena evaluates the labelCount query from line 269 of IndividualResponseBuilder.java:

    private static String LABEL_COUNT_QUERY = ""
        + "PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#> \n"
        + "SELECT ( str(COUNT(?label)) AS ?labelCount ) WHERE { \n"
        + "    ?subject rdfs:label ?label \n"
        + "    FILTER isLiteral(?label) \n"
        + "}" ;

Jena does not include a string type for the returned ?labelCount string. MarkLogic's REST API, on the other hand, when it evaluates that SPARQL query, returns ?labelCount as a typed literal with a string type annotation. In particular, it returns XML like this:

```xml
<sparql>
  <head>
    <variable name="labelCount" />
  </head>
  <results>
    <result>
      <binding name="labelCount">
        <literal datatype="http://www.w3.org/2001/XMLSchema#string">1515</literal>
      </binding>
    </result>
  </results>
 </sparql>
```

The Vitro code I propose changing implicitly relied on an untyped literal being returned. When a typed literal becomes an [RDFNode](https://jena.apache.org/documentation/javadoc/jena/org/apache/jena/rdf/model/RDFNode.html) and toString() is called, the resulting string representation includes the type annotation, so it cannot be parsed to an int. My crude fix uses a regex to remove the type annotation.

Another possible fix would be to eliminate the str() in the SPARQL query above, so the count is always typed as a number. Then the consuming code could convert it to an int with RDFNode.toLiteral().getInt() - but I'm not sure if there was a reason to convert the count to a string in the SPARQL query.

j2blake:
Thanks for the explanation, Patrick. I'll look at it.

patrickmcelwee:
Graham,

Yes, I think it would be (if there doesn't seem to be a reason to convert the count to a string in the SPARQL query ... and I can't think what the reason would be). If you agree, I'd be happy to update this pull request.

Thanks and welcome aboard!

grahamtriggs:
Thanks Patrick,

Please go ahead and update the pull request.

grahamtriggs:
Hi Patrick,

Closing this request as I've added the asLiteral / getInt change to the code.

Thanks for highlighting this!

patrickmcelwee:
Thanks Graham! Sorry I lost track of this.

PR #23, opened 2015-08-17
simplified getConfigFromSession logic with ternary operator

zednis:


j2blake:
Stephan - Thanks for the pull request. I'll merge it in.

In the future, it might be better to avoid pull requests that are purely stylistic in nature, since they are so subjective. For example, some instructors teach that the ternary operator is inherently less clear than if/then/else, and should never be used. Like you, I disagree with that. 

As another example, I try to avoid "negative logic", so I would have written this
```return (configs != null) ? configs.get( editKey ) : null;```
as this
```return (configs == null) ? null : configs.get( editKey );```
This avoids the verbalization, *"else, if configs is not not equal to null..."*. Again, it is a matter of preference, and open to debate.

And of course, our soon-to-be tech lead may have different opinions.

I would have enjoyed seeing you at the conference. - Jim

zednis:
Thanks Jim,

I'll avoid primarily stylistic change requests in the future.  I personally believe refactorings like this are more than just style changes, as they remove unnecessary conditional checks and make complicated code less error-prone (though there was no error that I was fixing in this case).  This was primarily a case of being able to refactor 7 lines into 1 very succinct statement.

- Stephan

j2blake:
Stephan - I understand, and you will see me doing a lot of this sort of refactoring. As a rule, I wait until I am already making changes to that section of code for some other reason.

zednis:
That is a sensible rule.  I was looking at a customized-version of the class in our 3-tier build and made the change there with some other changes in our customized version of the class.

j2blake:
Always a judgement call. Thanks for caring enough to improve the code base.

PR #24, opened 2015-11-12
Add YASQE - allows for syntax highlighting and prefix autocompletion.

lawlesst:
This adds the [YASQE](http://yasqe.yasgui.org/) SPARQL query editor to the default site admin SPARQL query tool/form. YASQE has many nice features like syntax highlighting and auto-completion. See image below for example. 

There's more that could be done with YASQE but this adds initial support without changing the functionality of the existing SPARQL admin form.

image

### DIFF
--- a/webapp/web/css/edit.css
+++ b/webapp/web/css/edit.css
@@ -12,9 +12,11 @@ div.sparqlform label {
 input.block {
     display: block;
 }
-textarea,
-div.sparqlform div {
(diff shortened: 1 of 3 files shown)


grahamtriggs:
Thanks Ted - looks interesting.

I've created a ticket for it to go into v1.9 (https://jira.duraspace.org/browse/VIVO-1217).

PR #25, opened 2015-12-02
tweak index page initHeight and don't add (cont'd) text if there aren't any more classes to show

gneissone:
The index page may roll over into 4 columns with the initHeight set to the container height / 2.1 (see attached). Adjusting to / 2.05 appears to fix the problem.
Additionally, added a fix that accounts for if the number of populated classes in the group was exactly equal to 35. Previously, the class group would be split even if there weren't any more classes to display.

screen shot 2015-12-01 at 11 09 08 am


grahamtriggs:
Thanks Benjamin - it was something I noticed ages ago in a previous job, but completely slipped my mind to look at it for v1.8.1 until I had committed to the release.

Glad to see a fix for this!

PR #28, opened 2016-01-08
bug fix for search help back link [VIVO-1226]

gneissone:
https://jira.duraspace.org/browse/VIVO-1226
Switched the Advanced Tips to be a div to avoid adding the anchor link to the browser history.


grahamtriggs:
Hi,

I've committed an alternative version of this fix.

It's better to keep Javascript separate rather than being inline attributes in the html. Also, it's arguably a problem to do this if a user has Javascript disabled, as it may mean they won't be able to see the advanced tips.

Instead, I've added to return false; statements to the Javascript block that adds the onclick functionality. This prevents the link being followed if the Javascript is executed (having the same result in not polluting the history), and if Javascript is disabled, the link will work normally.

I've also added the same fix to the Spanish template.

G

gneissone:
Excellent, works for me.

PR #29, opened 2016-01-08
Adds UTF-8 charset to emails

ardamose123:
This is required for emails that contain non-ASCII characters (like Spanish accents).

If *setContent(…)* has a MIME type but no charset, emails are sent with the given MIME type and UTF-8 charset, but the contents will not be encoded using UTF-8.

If an explicit charset is given, the emails are encoded correctly.


grahamtriggs:
Hi,

Thanks for the improvement!

There aren't currently any plans to make further releases to the 1.8.x line, so I have cherry picked your commits on to the develop branch, to be included in the next release (likely 1.9.0).

I've added a Jira ticket to document that the code has been included:
https://jira.duraspace.org/browse/VIVO-1228

Regards,
G

PR #38, opened 2016-05-25
[VIVO-1245] Reduce code duplication and class overwriting in VIVO

grahamtriggs:
This pull request changes the way VIVO customisations are applied to the dozen or so classes that were defined in Vitro and overlaid in the VIVO layer.

It's not "ideal" - in that it has extensions points that any application overlaying Vitro can perform customisations without adding extra hooks in the Vitro layer, it only adds extension points for what is currently used.

It does allow us to simplify the build process, and make the VIVO API artefact simply one that depends on Vitro API, without shading it - and therefore allowing Maven dependencies and IDEs autocompletion to work correctly.


j2blake:
These changes may affect any site with local modifications or a 3-tier build. Will 3-tier builds still work? I believe this is a good direction to go, but it's important that the upgrade notes are clear.
Also, if the "shading" is to be removed, it's important that the customization pages in the wiki are not left with obsolete "shading" instructions.

grahamtriggs:
Thanks Jim - my plan is to remove "shading" at the JAR (vitro-api / vivo-api) level. However, it would still be possible to "shade" on the class path by placing your Java classes directly in the Web App project (so that they are built to the "classes" folder rather than a jar in lib.

So three tier is still supported - but if you introduce another jar project, you should be careful not to "shade" existing classes (and only do that via the web app if you really need to).

Agreed, documentation is important - both for explaining the changes to the code / configuration files (which probably won't impact too many people directly), but also in how they adapt their traditional build environments to the Maven structure.

PR #46, opened 2016-06-28
update containsNullOrEmpty() to check if list is empty

zednis:
Also updated the method to use Apache Commons StringUtil::isBlank(String) function to test for blank values in the list.

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/EditN3GeneratorVTwo.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/EditN3GeneratorVTwo.java
@@ -11,6 +11,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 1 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/EditN3GeneratorVTwo.java
```diff
@@ -78,8 +79,14 @@ public void subInMultiUris(Map<String,List<String>> varsToVals, List<String> n3t
         }         
     }
   
-    private boolean containsNullOrEmpty(List<String> values) {        
-        return values != null && ( values.contains(null) || values.contains("") );       
+    private boolean containsNullOrEmpty(List<String> values) {
+    	//return values != null && ( values.contains(null) || values.contains("") );
```
nateprewitt: This may be a bit knitpicky but we can probably remove this line before merging.

zednis: Agreed.  I was also thinking of combining the values == null and values.isEmpty check.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/EditN3GeneratorVTwo.java
```diff
@@ -78,8 +79,14 @@ public void subInMultiUris(Map<String,List<String>> varsToVals, List<String> n3t
...
+    	if(values == null) return true ;
+    	if(values.isEmpty()) return true ;
```
nateprewitt: Another minor note, any reason not to merge this into a single check?

`if(values == null || values.isEmpty()) return true ;`

zednis: See my comment above ;-)  I was already thinking about it.  Sometimes I try to not over-collapse conditionals like that for improved readability and to make it slightly easier to comment out conditions during testing.  Functionally they are identical so there is no technical reason not to combine them.

zednis: I don't think this is an example of the combined conditional being overly complicated so I will update the proposed change to combine the conditionals.

nateprewitt:
@grahamtriggs this looks good to me. +1

grahamtriggs:
Thanks guys!

PR #48, opened 2016-07-06
[VIVO-1251] Special chars in property group names

rjyounes:
Allow special characters comma, forward slash, and ampersand in property group labels. Previously caused html error when the property group name was transformed into html attributes.


rjyounes:
Withdrawing this pull request - will add to develop branch

PR #57, opened 2017-04-04
[VIVO-1315] Plum print integration

smmckay:
https://jira.duraspace.org/browse/VIVO-1315

This is a VIVO-specific change but I went with the way that Altmetric is integrated.


grahamtriggs:
Hi,

I've integrated the code into develop, but I've set it to be disabled by default, as the spacing of the PlumX logo makes the list look odd. (unlike the AltMetric badge, which fits in a single line spacing).

If there is an update to the visual display, I'm happy to reconsider it being enabled by default, but for now, I'll leave it for people to enable if they want to have it.

Thanks,
Graham

PR #61, opened 2017-07-26
Enable Plum print by default

smmckay:
Now that the Plum print in an author's publication list is small enough not to break the line height, maybe it can be enabled by default.


mjaved495:
Graham, does that mean we will have altmetric and plum metric for a publication ?

grahamtriggs:
Hi Javed,

It has been an option for a while. This just makes it the default to have both, but you can disable either (or both) in the runtime.properties.

PR #63, opened 2017-12-15
Look in vitro.home/config directory for runtime.properties at startup

gneissone:
The application will look in vitro.home/config for runtime.properties if it doesn't find it in vitro.home. example.runtime.properties is located in the config directory so seems like a logical place to look for runtime.properties.

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/config/ConfigurationPropertiesSetup.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/config/ConfigurationPropertiesSetup.java
@@ -23,22 +23,22 @@ import edu.cornell.mannlib.vitro.webapp.startup.StartupStatus;
 /**
  * Locates the runtime configuration properties and stores them in the servlet
  * context.
- * 
+ *
  * This must be invoked before any listener that requires configuration
  * properties.
- * 
+ *
  * The properties are determined from a file called 'build.properties' in the
  * resources directory of the webapp, and a file called 'runtime.properties' in
  * the Vitro home directory. In case of conflict, runtime.properties wins.
- * 
+ *
  * The path to the Vitro home directory can be specifed by an JNDI value, or by
  * a System property, or by a property in build.properties, in that order. If
  * the Vitro home directory is specified in more than one way, a warning is
  * issued and the first value is used.
- * 
+ *
  * The value that was determined for 'vitro.home' is also included in the
  * ConfigurationProperties bean.
- * 
+ *
  * If build.properties or runtime.properties cannot be located or loaded, a
  * fatal error is registered to abort the startup.
  */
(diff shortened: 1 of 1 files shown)


gneissone:
@vivo-project/vivo-committers @mconlon17 does anybody else have a thought on this PR? It's a small change, just look in vitro.home/config for runtime.properties if it doesn't exist in vitro.home. Does there need to be a note in runtime.properties saying that if there are runtime.properties files in both directories that the one in vitro.home takes precedence? VIVO's example.runtime.properties currently says 
`# Save a copy of this file as runtime.properties in your VIVO home directory,`  
`# and edit the properties as needed for your installation. `

mconlon17:
I can leave my runtime.properties in vivo-home/config?  I'm all for that.  The above says vitro-home.

??

gneissone:
Yes, these changes allow you to leave runtime.properties in vivo-home/config instead of copying them over to vivo-home. 

Vitro home and VIVO home are the same thing in practice. Vitro home is overridden by VIVO home thanks to VIVO's [context.xml](https://github.com/vivo-project/VIVO/blob/47f1408b3e9db4934c33f321689ef307b9cf4bdd/installer/webapp/src/main/webResources/META-INF/context.xml).

gneissone:
VIVO now raises a deprecation warning on startup if runtime.properties is found in vivo.home instead of vivo.home/config or if it's found in both. Via ConfigurationPropertiesSmokeTests.

fereira:
I didn't get a chance to deploy this but from a code review it looks good to me.

j2blake:
Benjamin, will you be updating https://wiki.duraspace.org/display/VIVODOC110x/Installing+VIVO#InstallingVIVO-ConfiguretheHomeDirectory and https://wiki.duraspace.org/display/VIVODOC110x/Upgrading+VIVO to reflect this change?

gneissone:
Will make a note to edit the wiki if this is merged.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/config/ConfigurationPropertiesSetup.java
```diff
@@ -52,6 +52,9 @@
 	/** Configuration property to store the Vitro home directory */
 	private static final String VHD_CONFIGURATION_PROPERTY = "vitro.home";
 
+	/** Configuration property used to determine if there are runtime.properties files in multiple locations **/
+	private static final String RP_MULTIPLE = "rp.multiple";
```
j2blake: Why not give this package-protection (default) instead of private. Then you could refer to it from ConfigurationPropertiesSmokeTests, instead of declaring it a second time.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/config/ConfigurationPropertiesSmokeTests.java
```diff
@@ -72,9 +74,35 @@ private void checkDefaultNamespace(ServletContext ctx,
...
 
+	/**
+	 * Warn if runtime.properties exists in multiple locations
+	 * or is located vivo.home instead of vivo.home/config
+	 */
+	private void checkMultipleRPFs(ServletContext ctx,
+			ConfigurationProperties props, StartupStatus ss) {
+		String rpfStatus = props.getProperty(MULTIPLE_RPFS);
+
+		if (rpfStatus == "both") {
```
j2blake: .equals() would be more robust than == 
This will work because you used an interned string in ConfigurationPropertiesSetup.

PR #71, opened 2018-04-26
VIVO-1475 Add Data Distribution API

j2blake:
**[JIRA Issue](https://jira.duraspace.org/projects/VIVO)**: https://jira.duraspace.org/browse/VIVO-1475

* Other Relevant Links 
    * [API Documentation](https://cul-it.github.io/vivo-data-distribution-api/index.html)

Adds the Data Distribution API (developed by Scholars@Cornell) to the VIVO distribution.

# What's new?
Allows the site administrator to configure data distributors - named, parameterized requests for data from VIVO.

# How should this be tested?
1. Bare API:
    * Install and start VIVO
    * Point browser to `http://localhost:8080/vivo/api/dataRequest`
        * Response is `'action' path was not provided.`
    * Point browser to `http://localhost:8080/vivo/api/dataRequest/hello?name=World`
        * Response is `Did not find a DataDistributor for 'hello'`
2. Configure a distributor:
    * Create a configuration file, as described in [Using in VIVO 1.10](https://cul-it.github.io/vivo-data-distribution-api/install_vivo_1_10.html)
    * Install and start VIVO
    * Point browser to `http://localhost:8080/vivo/api/dataRequest/hello?name=World`
        * Response is `Hello, World!`

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -252,5 +252,11 @@
             <artifactId>javax.mail</artifactId>
             <version>1.6.0</version>
         </dependency>
+        
+        <dependency>
+            <groupId>edu.cornell.library.scholars</groupId>
+            <artifactId>data-distribution-api-vivo_1_10</artifactId>
+            <version>1.1.1</version>
+        </dependency>
     </dependencies>
 </project>


awoods:
I believe the starting assumption is that the Data Distribution API is functionality that we want to make generally available to the community.
The question becomes, "how do we enable that in a way that does not overly couple dependencies."
The ideal scenario may be one where:
1. A default DDAPI comes with the VIVO release
1. The version of the DDAPI can be updated on an independent schedule from VIVO releases
1. The inclusion of the DDAPI could be excluded from a given site's installation

Assuming the DDAPI functionality does come into the VIVO release, there is a separate line of discussion to be had related to redesigning visualizations to use the DDAPI.

Let's discuss in detail during next Wednesday's (May 2, 2018) [Committer call](https://docs.google.com/document/d/1mRjN5O4lCLHjtUDyYhCje1YpuAVaZuZsvGiBw_fZUb8/edit?usp=sharing).

mconlon17:
My understanding is that inclusion provides no functionality without additional configuration by the site.

awoods:
That is my understanding as well, @mconlon17 . 
However, the dependency is included, which relates to @grahamtriggs' points.

mjaved495:
1. As name says, "Data Distribution" API is not to provide data for visualizations only, but any data can be accessible via DD API, in many formats (JSON, CSV, RDF etc.). Currently, BTI (an institution at Cornell) access Publications data in JSON using DD API.
2. I feel, the maven term "dependency" is somewhat misleading here. VIVO application is not dependent on DD API. By adding it into the application (as dependency), it provided additional functionalities and enhancements that currently does not exist in VIVO. 
3. When (in future) we redesign current visualizations to use the DDAPI, at that time, DD API would be a real "dependency" for VIVO application and we may want to include it in the VIVO core at that time.

grahamtriggs:
In my mind, there is a clear distinction between functionality that is provided "out of the box" (even if it requires further configuration to be useful), and things that a user chooses to add for themselves.
There is also a clear distinction between what is documented as part of a release, and what is documented elsewhere.
If the code is going to be present in the application without the user having done anything, then it is "out of the box". And if that is the case, and/or it is documented within the release documentation, then it is creating an expectation for future support.
If this is included in a pom - even as a commented out entry for easy inclusion (like the languages) - then it creates an interesting chicken-and-egg situation. You can't readily release the DD artifact without the release of Vitro/VIVO, And you can't release the Vitro/VIVO without (at least knowing, in the case of a comment), the version of the DD artifact.

Note also that the "dependencies" project is not meant to be a place to put all Maven dependencies. It was a solution required initially to describe the dependencies required for the API project, which could then also be supplied to the webapps (where we could not use transitive dependencies). The problems that led to that approach (VIVO api was an overlay to Vitro API) have been resolved, so we could move them to the API project and remove the dependencies project altogether.

The only practical resolutions I see is that either the code is included in, and released as part of, Vitro, or to have this as an after-market add-on where people are instructed how to add the dependency to their installer/webapp/pom.xml.

awoods:
Tangent: @grahamtriggs, would you mind creating a JIRA ticket (and working it if you like) for removing the https://github.com/vivo-project/dependencies project?

mconlon17:
Built.  Tested using Jim's suggestions above.  Works.

fereira:
From our discussion in the committers call.

The DistributeDataApiController servlet that needs to be included in the web.xml file does not use annotations (other than @Override).  

The ExtendedRDF feature I mentioned is actually called "rich-export" and actually deprecated in vivo version 1.8 (and is still just marked deprecated).  Adding "serveExtendedLinkedData=true" to runtime properties will turn it on, then the IndividualControlller will use a deprecated ExtendedRdfAssembler class to generate RDF.  That class will look in WEB-INF/rich-export directory for sparql files which construct the "extended linked data".  All those files still exist in the develop branch.  I've used the feature on an old vivo instance but don't know if anyone else is actually using it.

PR #73, opened 2018-05-09
Moved password encryption from MD5 to a salted and secure hash - 1448

Asimq:
https://jira.duraspace.org/browse/VIVO-1448

**What does this pull request do?**

It replaces the MD5 password hashing with a more secure and salted password hashing algorithm.

**What's new?**

MD5 hashing has been replaced by Argon2i hashing, which is salted and configurable in terms of memory, time and parallelism needed to compute a hash. https://github.com/phxql/argon2-jvm

MD5 hashing has not been completely eliminated and the users with MD5 hash as their passwords are asked by the system to renew their password once they log in.

The configurable parameters of the Argon2i algorithm are stored in the "runtime.properties" file.

**How should this be tested?**

- Copy the useraccounts (with Md5 passwords) from an already installed instance of Vivo/Vitro and place them in the home directory of vitro with Argon2i hashing and check if the system promotes you to renew your password after logging in.
- Login using API with root credentials.  (Could be done using a browser i.e, http://[server address e.g., localhost:8080]/[application name e.g., vitro]/api/sparqlQuery?email=[root email]&password=[root password])
- Create a new user and login
- Change your password and login
- Create an invite/user/create-user link
- Check external authentical
- Change the values for parallelism, memory and time in runtime.properties and login ( you should be still able to login but the time required to login should change)
- Restrict user accounts and login

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -54,6 +54,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>de.mkammerer</groupId>
+            <artifactId>argon2-jvm</artifactId>
+            <version>2.4</version>
+        </dependency>
+        <dependency>
             <groupId>org.vivoweb</groupId>
             <artifactId>vitro-dependencies</artifactId>
             <version>2.0.0-SNAPSHOT</version>
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/auth/policy/PolicyHelper.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/auth/policy/PolicyHelper.java
@@ -102,7 +102,7 @@ public class PolicyHelper {
 			String uri = user.getUri();
 			log.debug("userAccount is '" + uri + "'");
 
-			if (!auth.isCurrentPassword(user, password)) {
+			if (!auth.isCurrentPasswordArgon2(user, password)) {
 				log.debug(String.format("UNAUTHORIZED, password not accepted "
 						+ "for %s, account URI: %s", email, uri));
 				return false;
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/auth/policy/RootUserPolicy.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/auth/policy/RootUserPolicy.java
@@ -67,6 +67,7 @@ public class RootUserPolicy implements PolicyIface {
 		private ServletContext ctx;
 		private StartupStatus ss;
 		private UserAccountsDao uaDao;
+		private ConfigurationProperties cp;
 		private String configuredRootUser;
 		private boolean configuredRootUserExists;
 		private TreeSet<String> otherRootUsers;
@@ -75,6 +76,7 @@ public class RootUserPolicy implements PolicyIface {
 		public void contextInitialized(ServletContextEvent sce) {
 			ctx = sce.getServletContext();
 			ss = StartupStatus.getBean(ctx);
+			cp = ConfigurationProperties.getBean(ctx);
 
 			try {
 				uaDao = ModelAccess.on(ctx).getWebappDaoFactory()
@@ -148,8 +150,8 @@ public class RootUserPolicy implements PolicyIface {
 			ua.setEmailAddress(configuredRootUser);
 			ua.setFirstName("root");
 			ua.setLastName("user");
-			ua.setMd5Password(Authenticator
-					.applyMd5Encoding(ROOT_USER_INITIAL_PASSWORD));
+			ua.setArgon2Password(Authenticator.applyArgon2iEncoding(cp,ROOT_USER_INITIAL_PASSWORD));
+			ua.setMd5Password("");
 			ua.setPasswordChangeRequired(true);
 			ua.setStatus(Status.ACTIVE);
 			ua.setRootUser(true);
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/UserAccount.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/UserAccount.java
@@ -48,6 +48,7 @@ public class UserAccount {
 	private String firstName = ""; // Never null.
 	private String lastName = ""; // Never null.
 
+	private String argon2Password = ""; //Never null.
 	private String md5Password = ""; // Never null.
 	private String oldPassword = ""; // Never null.
 	private long passwordLinkExpires = 0L; // Never negative.
@@ -104,6 +105,14 @@ public class UserAccount {
 		this.lastName = nonNull(lastName, "");
 	}
 
+	public String getArgon2Password() {
+		return argon2Password;
+	}
+
+	public void setArgon2Password(String argo2Password) {
+		this.argon2Password = nonNull(argo2Password, "");
+	}
+
 	public String getMd5Password() {
 		return md5Password;
 	}
@@ -125,8 +134,9 @@ public class UserAccount {
 	}
 
 	public String getPasswordLinkExpiresHash() {
-		return limitStringLength(8, Authenticator.applyMd5Encoding(String
+		return limitStringLength(8, Authenticator.applyArgon2iEncoding(String
 				.valueOf(passwordLinkExpires)));
+		//applyMd5Encoding
 	}
 
 	public void setPasswordLinkExpires(long passwordLinkExpires) {
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/accounts/UserAccountsSelector.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/accounts/UserAccountsSelector.java
@@ -246,6 +246,7 @@ public class UserAccountsSelector {
 			user.setFirstName(ifLiteralPresent(solution, "firstName", ""));
 			user.setLastName(ifLiteralPresent(solution, "lastName", ""));
 			user.setMd5Password(ifLiteralPresent(solution, "pwd", ""));
+			user.setArgon2Password(ifLiteralPresent(solution, "pwd", ""));
 			user.setPasswordLinkExpires(ifLongPresent(solution, "expire", 0L));
 			user.setLoginCount(ifIntPresent(solution, "count", 0));
 			user.setLastLoginTime(ifLongPresent(solution, "lastLogin", 0));
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/accounts/admin/UserAccountsAddPageStrategy.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/accounts/admin/UserAccountsAddPageStrategy.java
@@ -198,8 +198,8 @@ public abstract class UserAccountsAddPageStrategy extends UserAccountsPage {
 		@Override
 		protected void setAdditionalProperties(UserAccount u) {
 			if (!page.isExternalAuthOnly()) {
-				u.setMd5Password(Authenticator
-						.applyMd5Encoding(initialPassword));
+				u.setArgon2Password(Authenticator.applyArgon2iEncoding(initialPassword));
+				u.setMd5Password("");
 				u.setPasswordChangeRequired(true);
 			}
 			u.setStatus(Status.ACTIVE);
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/accounts/admin/UserAccountsEditPageStrategy.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/accounts/admin/UserAccountsEditPageStrategy.java
@@ -194,7 +194,8 @@ public abstract class UserAccountsEditPageStrategy extends UserAccountsPage {
 		@Override
 		protected void setAdditionalProperties(UserAccount u) {
 			if (!page.isExternalAuthOnly() && !newPassword.isEmpty()) {
-				u.setMd5Password(Authenticator.applyMd5Encoding(newPassword));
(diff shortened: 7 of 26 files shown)


KitioFofack:
I just reviewed this PR.
@Asimq  just need to:
 
1. to update the version in Vitro to 1.10.SNAPSHOT
2. add the argon configuration in example.runtime.properties of VIVO

Asimq:
@KitioFofack I've updated the Snapshot versions to 1.10.0 from 2.0.0.
I couldn't find example.runtime.properties in Vivo as it's a part of Vitro.

j2blake:
@Asimq - please create a pull request which makes these changes to example.runtime.properties in VIVO.

j2blake:
# Summary 
I have examined the code changes. I have tested:

* normal logins, 
* interrupted logins, 
* failed login attempts,
* login via the `/programLogin` and `/admin/login` pages,
* login to the SPARQL Update API.

It all works as intended (after changes suggested by Kitio). The code is essentially clean and functional.
I have suggestions for stylistic and structural changes, listed below.
I also suggest some additions to unit tests and smoke tests.

I have two major areas of concern:

* There should be an accompanying pull request for VIVO, putting the Argon2 settings into `runtime.properties` (Kitio has already pointed this out).
* The user experience is a problem. When an old system is upgraded, each user must enter a new password when they log in. No explanation is given. The user has no idea why they are being asked to enter a new password.
This seems serious enough to delay this pull request. I will ask the developer's channel for comment.

# Requested changes

## Code style
VIVO has no official style guide for code. However, if a class has a consistent coding style, modifications to that class should copy that style.

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/auth/policy/RootUserPolicy.java`
* Use standard Java coding style guidelines, like space after comma.
* Wrap code to approx 80 columns.

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/UserAccount.java`
* Remove obsolete and misleading comment, line 139
* Add a line to the toString() method, to include the argon2Password in the output, after line 247:

	````
				+ (", argon2password=" + argon2Password)
	````

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/Authenticator.java`
* Line 62, fix the indentation
* What does `isCurrentPasswordArgon2()` do? Add a documentation comment.
* In `applyArgon2iEncoding()`, (both forms) wrap the code to 80 columns, 
  use braces on then and else clauses, fix the indentation.

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/BasicAuthenticator.java`
* Coding style: line wrapping and space after comma

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/ProgramLogin.java`
* Remove commented code (or add comment explaining why it is there)

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/RestrictedAuthenticator.java`
* Coding style: line wrapping and space after comma

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/edit/Authenticate.java`
* Indentation is inconsistent.
* Should have the same braces style that is used in the surrounding code.

#### `api/src/test/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/AuthenticatorStub.java`
* Coding style: line wrapping and space after comma

#### `api/src/test/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/ProgramLoginTest.java`
* Remove commented code (or add comment explaining why it is there)
* Remove `System.out.println()` statements. Unit tests should run silently except for pass/fail result.

#### `api/src/test/java/edu/cornell/mannlib/vitro/webapp/controller/edit/AuthenticateTest.java`
* Remove commented code (or add comment explaining why it is there)

#### `api/src/test/java/edu/cornell/mannlib/vitro/webapp/dao/jena/UserAccountsDaoJenaTest.java`
* Remove commented code (or add comment explaining why it is there)

#### `home/src/main/resources/config/example.runtime.properties`
* Wrap comments to approx 80 columns.

#### `legacy/webapp/config/example.runtime.properties`
* Wrap comments to approx 80 columns.


## Code structure

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/Authenticator.java`
* In `applyArgon2iEncoding()`, (both forms), why use a try/catch? 
  If the parameters are missing from ConfigurationProperties, you will throw a RuntimeException, 
  catch it, and wrap it in another RuntimeException. What is the reason?

* I don't like having two versions of `applyArgon2iEncoding()`. Certainly not with this "cut and paste" style of duplicate code.
	* Consider this:
	
	````
	public static String applyArgon2iEncoding(String raw) {
		Argon2 argon2 = Argon2Factory.create();
		if (cp.getProperty("argon2.time") != null
				&& cp.getProperty("argon2.memory") != null
				&& cp.getProperty("argon2.parallelism") != null) {
			return argon2.hash(Integer.parseInt(cp.getProperty("argon2.time")),
					Integer.parseInt(cp.getProperty("argon2.memory")),
					Integer.parseInt(cp.getProperty("argon2.parallelism")),
					raw);
		} else {
			throw new RuntimeException(
					"Parameters \"argon2.time\", \"argon2.memory\" and "
							+ "\"argon2.parallelism\" are either missing in the "
							+ "\"runtime.properties\" file or are not defined correctly");
		}
	}

	public static String applyArgon2iEncoding(ConfigurationProperties configProperties, String raw) {
		cp = configProperties;
		return applyArgon2iEncoding(raw);
	}
	````
	
	* If it feels uncomfortable to set a new value into `cp`, we could do it the other way around:
	
	````
	public static String applyArgon2iEncoding(String raw) {
		return applyArgon2iEncoding(cp, raw);
	}

	public static String applyArgon2iEncoding(
			ConfigurationProperties configProp, String raw) {
		Argon2 argon2 = Argon2Factory.create();
		if (configProp.getProperty("argon2.time") != null
				&& configProp.getProperty("argon2.memory") != null
				&& configProp.getProperty("argon2.parallelism") != null) {
			return argon2.hash(
					Integer.parseInt(configProp.getProperty("argon2.time")),
					Integer.parseInt(configProp.getProperty("argon2.memory")),
					Integer.parseInt(configProp.getProperty("argon2.parallelism")), raw);
		} else {
			throw new RuntimeException(
					"Parameters \"argon2.time\", \"argon2.memory\" and "
					+ "\"argon2.parallelism\" are either missing in the "
					+ "\"runtime.properties\" file or are not defined correctly");
		}
	}
	````
	
	* If we can't rely on `cp` having a valid value, perhaps this is a sign 
	  that `cp` shouldn't be a static variable to begin with. 
	  There is another option which requires only one method, and no static variable.
	
	````
	public static String applyArgon2iEncoding(String raw) {
		ServletContext ctx = ApplicationUtils.instance().getServletContext();
		ConfigurationProperties configProp = ConfigurationProperties.getBean(ctx);

		Argon2 argon2 = Argon2Factory.create();
		if (configProp.getProperty("argon2.time") != null
				&& configProp.getProperty("argon2.memory") != null
				&& configProp.getProperty("argon2.parallelism") != null) {
			return argon2.hash(
					Integer.parseInt(configProp.getProperty("argon2.time")),
					Integer.parseInt(configProp.getProperty("argon2.memory")),
					Integer.parseInt(configProp.getProperty("argon2.parallelism")), raw);
		} else {
			throw new RuntimeException(
					"Parameters \"argon2.time\", \"argon2.memory\" and "
							+ "\"argon2.parallelism\" are either missing in the "
							+ "\"runtime.properties\" file or are not defined correctly");
	   }
	}
	````
	
	If you go this way, don't forget to remove the static `cp` variable.
	
* The method `applyArgon2iEncodingStub()` does not belong here, since it is only used in the tests. 
  Why not add it as a static method on `AuthenticatorStub`, instead? 
  No need to muddy the live code with test methods.


## Logic

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/accounts/UserAccountsSelector.java`
* _**WARNING:**_ As written, this will set both MD5 password and Argon2 password fields to the same value:
  the stored value for MD5 password. Here is one solution:
	* Change line 39-40 to read:
	
	````
	private static final String ALL_VARIABLES = "?uri ?email ?firstName "
			+ "?lastName ?md5pwd ?a2pwd ?expire ?count ?lastLogin ?status ?isRoot";
	````
	* Replace line 161 with these two lines:
	
	````
				+ "    OPTIONAL { ?uri auth:md5password ?md5pwd } \n"
				+ "    OPTIONAL { ?uri auth:argon2password ?a2pwd } \n"
	````
	* Change line 249-250 (previously 248-249) to:
	
	````
			user.setMd5Password(ifLiteralPresent(solution, "md5pwd", ""));
			user.setArgon2Password(ifLiteralPresent(solution, "a2pwd", ""));
	````
	* Verify with unit tests, see below.

## Unit tests

#### `api/src/test/java/edu/cornell/mannlib/vitro/webapp/dao/jena/UserAccountsDaoJenaTest.java`
* Why doesn't line 394 set the value that was provided in line 382?
* At least one of your test cases should have a value of md5Password that is 
  not "", to demonstrate that it gets set correctly.
  
#### `api/src/test/java/edu/cornell/mannlib/vitro/webapp/controller/accounts/UserAccountsSelectorTest.java`
* Add a minimal test that `UserAccountsSelector` is properly loading 
  both the md5 and argon2 passwords from storage.
  * Replace line 84 with these lines:
  
  ````
		assertEquals("md5password", "garbage", acct.getMd5Password());
		assertEquals("argon2password", "betterGarbage", acct.getArgon2Password());
  ````
  * Replace line 109 (previously was 108) with these lines:
  
  ````
		assertEquals("md5password", "garbage", acct.getMd5Password());
		assertEquals("argon2password", "betterGarbage", acct.getArgon2Password());
  ````
  * in `api/src/test/resources/edu/cornell/mannlib/vitro/webapp/controller/accounts/UserAccountsSelectorTest.n3`
, insert 
    this line after line 115 and again after line 141:
    
  ````
    auth:argon2password "betterGarbage" ;
  ````

## Smoke tests
The argon2 properties in runtime.properties are required. As written, however, VIVO starts successfully if these properties are missing. Then, when someone goes to log in, the server returns an HTTP 500 error. 

This is why the smoke tests are created: to fail at startup with an informative message, rather than a mysterious failure at some later point.

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/config/ConfigurationPropertiesSmokeTests.java`
* Suggest these changes:
	* Add after line 31:

	````
	private static final String PROPERTY_ARGON2_TIME = "argon2.time";
	private static final String PROPERTY_ARGON2_MEMORY = "argon2.memory";
	private static final String PROPERTY_ARGON2_PARALLELISM = "argon2.parallelism";
	````
	
	* Add after line 43 (was line 40):
	
	````
		checkEncryptionParameters(props, ss);
	````
	
	* Add after line 129 (was line 125):

	````
	/**
	 * Fail if there are no config properties for the Argon2 encryption.
	 */
	private void checkEncryptionParameters(ConfigurationProperties props,
			StartupStatus ss) {
		failIfNotPresent(props, ss, PROPERTY_ARGON2_TIME);
		failIfNotPresent(props, ss, PROPERTY_ARGON2_MEMORY);
		failIfNotPresent(props, ss, PROPERTY_ARGON2_PARALLELISM);
	}

	private void failIfNotPresent(ConfigurationProperties props,
			StartupStatus ss, String name) {
		String value = props.getProperty(name);
		if (value == null || value.isEmpty()) {
			ss.fatal(this, "runtime.properties does not contain a value for '"
					+ name + "'");
			return;
		}
	}
	````

j2blake:
Here is one way to fix the user experience issue:

#### `api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/edit/Authenticate.java`
* Add after line 352:

````
					bean.setMessage(request, ERROR, "password_system_has_changed");
````

#### `webapp/src/main/webapp/i18n/all.properties`
* Add after line 125:

````
password_system_has_changed = Our password system has been upgraded to make it more secure. Please enter a new password.
````

_**NOTE:**_ This message is only a suggestion. Please feel free to improve.

brianjlowe:
Apologies if this is a naive question, as this is surely far from my area of expertise, but is it strictly necessary to require users to enter new passwords just to support the new encryption standard?  I seem to recall reading somewhere that another approach is to cascade the hash functions -- in this case we would now use argon2(md5(password)) instead of just argon2(password).  Would this not allow us to a batch re-hash the existing md5 hashes in the database so that the existing passwords would continue to work?

Asimq:
@brianjlowe After the update, when a user would login he/she would be asked to set a new password. 
The new password would be hashed as argon2(password) instead of md5(password).

brianjlowe:
@Asimq Understood that this is the behavior as currently implemented.  My question was more one of whether this behavior is necessary and/or desired.  

Related question:  what is the "user experience" for scripted API calls where usernames and passwords are passed as URL parameters?

Asimq:
@brianjlowe  We dicussed the possibilities and this behavious was something that everyone agreed upon.
Regarding API, a "root" user can login using the API but there's no mechnaism to change password there.

j2blake:
Very interesting, @brianjlowe -- another way to deal with this, and it sounds like it would be quite satisfactory. Perhaps that approach should be a new JIRA issue, with a target release of 1.10.1. But this implementation seems good enough for the current release, don't you think?

I'm ready to put approve this as it stands.

j2blake:
@KitioFofack, you commented on this PR a few days ago, and it seemed like you were satisfied. Is that true? Can you click whatever button is required for approval?

KitioFofack:
The idea of @brianjlowe  is seducing me but I agree with you @j2blake what we have now is good enough to go in the RC.

Asimq:
@j2blake and @KitioFofack could you guys also approve https://github.com/vivo-project/VIVO/pull/70

KitioFofack:
@Asimq I think I did

PR #74, opened 2018-05-18
Add check for language files to smoke tests and remove references to …

gneissone:
…build.properties

**[JIRA Issue VIVO-1501](https://jira.duraspace.org/browse/VIVO-1501)**: ConfigurationPropertiesSmokeTests raises outdated warning on multi-language startup

This PR fixes a problem where a warning was raised during startup anytime you started VIVO and are using selectable internationalization.

# What's new?
Adds a test to the configuration smoke tests to check if language files exist at the top webapp level that correspond to all the selectable languages. If the correct language files don't exist, a warning is thrown during startup. 

Also removes a test that used to check if languages were specified in build.properties, which no longer exists and adjusts a test that checked for languages in build.properties to use the languages specified by selectableLanguages in runtime.properties.

# How should this be tested?
Replicate the issue by deploying VIVO with the vivo-languages package included and adding a couple languages to the selectable languages setting in runtime.properties. A warning is thrown about languages not being included in build.properties (which doesn't exist anymore).

Try again with this PR. Above warning should not appear.
Try adding a valid language that isn't included in vivo-languages (e.g. fr_FR) to selectableLanguages in runtime.properties and check that warning is raised during startup. You could also try an invalid locale (e.g. fo_BA). The same warning will be raised and you'll also get a warning from LocaleUtils that it's an invalid locale.

# Additional Notes:
Any additional information that you think would be helpful when reviewing this PR.

Example:
* Does this change require documentation to be updated? 
* Does this change add any new dependencies? 
* Does this change require any other modifications to be made to the repository? 
* Could this change impact execution of existing code?

# Interested parties
@KitioFofack


gneissone:
@j2blake suggestions to improve the Java are always appreciated, I hacked this together.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/config/ConfigurationPropertiesSmokeTests.java
```diff
@@ -119,14 +123,46 @@ private void checkLanguages(ConfigurationProperties props, StartupStatus ss) {
...
+		}
+
+		List<String> i18nNames = null;
+
+		try {
+		  i18nNames = geti18nNames(i18nDirPath);
+		}
+		catch(java.io.IOException e) {
+		  e.printStackTrace();
+		}
```
j2blake: `e.printStackTrace()` just dumps to the log, and not even to the VIVO log -- goes against the purpose of the smoke test. Handle the exception inside `geti18nNames()`

api/src/main/java/edu/cornell/mannlib/vitro/webapp/config/ConfigurationPropertiesSmokeTests.java
```diff
@@ -4,6 +4,12 @@
 
 import java.net.URI;
```
j2blake: Add another import:
````
import java.io.IOException;
````

File: api/src/main/java/edu/cornell/mannlib/vitro/webapp/config/ConfigurationPropertiesSmokeTests.java [hunk 3]
```diff
@@ -138,17 +174,29 @@ private void checkLanguages(ConfigurationProperties props, StartupStatus ss) {
...
+	private List<String> geti18nNames(String i18nBaseDirPath) throws java.io.IOException {
+		List<String> i18nNames = Files.walk(Paths.get(i18nBaseDirPath))
+				.filter(Files::isRegularFile)
+				.map(Path::getFileName)
+				.map(p -> {return p.toString();})
+		    .collect(Collectors.toList());
+
+		return i18nNames;
+	}
+
```
j2blake: Rather than pass the `IOException`, we can wrap it in a `RuntimeException`, so it will not be caught anywhere in this class. The `StartupManager` will catch it, and do the equivalent of `ss.fatal(this, "some simple message", e)`
````
	/** Create a list of the names of available language files. */
	private List<String> geti18nNames(String i18nBaseDirPath) {
		try {
			List<String> i18nNames = Files.walk(Paths.get(i18nBaseDirPath))
					.filter(Files::isRegularFile)
					.map(Path::getFileName)
					.map(p -> {return p.toString();})
					.collect(Collectors.toList());

			return i18nNames;
		} catch (IOException e) {
			throw new RuntimeException("Failed to find language files", e);
		}
	}
````

gneissone:
@j2blake thanks a bunch for the help, I have committed your suggestions. In the future, feel free to commit your improvements directly to the branch if you want. I'm submitting PRs based off branches of the main Vitro repo, now.

j2blake:
OK, Ben, I’ll keep that in mind.

Jim

gneissone:
@hudajkhan I believe I have addressed the problem by putting the check within an if statement, please try again. Additionally, I noticed a discrepancy between what runtime.properties said was the default value of RDFService.languageFilter (false) and what the smoke tests assumed (true). Dug into the configuration and determined that the correct value was in fact false, so switched it up in the smoke tests.

PR #76, opened 2018-06-27
[VIVO-1528] Select an editor for input fields based on annotation

Stefan-Wolff:
**[JIRA Issue](https://jira.duraspace.org/projects/VIVO)**: https://jira.duraspace.org/browse/VIVO-1528

It selects an editor based on the datatype annotation. If there is no such annotation, then the editor will be selected based an the range of the property to be edited.

# What's new?
WYSIWYG editor will be provided for data properties annotated with vitro:editing="HTML":
vivo:overview, vivo:teachingOverview, vivo:researchOverview, vivo:outreachOverview

# How should this be tested?

1. Edit a property having a range of integer, int, float, boolean, URI/URL or all types of date/time -> same behavior as before
2. Edit a property having a range of string or untyped and NOT having the annotation vitro:editing="HTML" -> input field should be provided (this is the new behavior)
3. Edit a property having the annotation vitro:editing="HTML" -> WYSIWYG editor should be provided

# Notice
1. You have to generate a firsttime start-up condition to apply the new annotations.
2. This pull request includes https://github.com/vivo-project/VIVO/pull/100 .


gneissone:
screen shot 2018-10-19 at 9 38 52 am

I'm guessing that the HTML type needs added to property forms, and the XML type needs edited?

gneissone:
I agree with Huda, some data properties would benefit from a larger input text input. Overviews and abstracts come to mind. I think this is the exception rather than the rule, so I'm okay with the default being this small form. 

Properties that require the WYSIWYG form could have their range changed to HTML in vivo.owl. @mconlon17 is the ontology group okay with this? The ranges we declare in vivo.owl don't appear to align exactly with the ontologies upstream (e.g. the range for vcard:hasMember in the vCard ontology is vcard:Kind, but it is vcard:Organization, vcard:Group, and vcard:Individual in vivo.owl). Alternatively, we can change the input form for those properties that need to WYWIWYG editor to use a (new?) custom form instead of this default, which I believe can be set in PropertyConfig.n3.

mconlon17:
As Huda noted above, The ontology group is fine with having the range on some datatype properties changed to HTML.  In particular vivo:overview (this should be required as a result of merging this pull request -- overviews often contain HTML and without the corresponding change, this pull request would prevent overviews from being edited WYSIWYG.

mconlon17:
@grahamtriggs has raised concerns about the approach here.  The rdf:HTML datatype can not be assigned a language tag.  Only the langString can be assigned a language tag, and in RDF 1.1, any literal with a language tag is assumed to be rdf:langString.  The approach proposed in this pull request of typing properties for the purpose of displaying an appropriate editor might better be served by using an annotation rather than a datatype.  This would allow the datatype to be langString where necessary and still provide WYSIWIG editing where appropriate.  Unfortunately, I don't know if we currenty have the annotation capability needed for Graham's approach.  Perhaps Graham can comment.

If we do not, we could greatly improve the user experience and data quality by merging this pull request, and assigning rdf:HTML to some properties.  This might be considered temporary until we are able to develop the annotation approach.

awoods:
> If we do not have annotation for editor, we could greatly improve the user experience and data quality by merging this pull request, and assigning rdf:HTML to some properties. This might be considered temporary until we are able to develop the annotation approach.

Pending @grahamtriggs 's response to the annotation question, I would support the incremental (temporary) improvement provided by this pull-request.

grahamtriggs:
The problem here is it isn't just a case of assigning a different type to some properties in the ontology - it changes the way that (new) instances of those properties are recorded in the triple store. But it won't automatically change anything that is already recorded in the content store.

It then raises questions about how the edit form works - will it display a WYSIWYG on the basis of the ontology definition, or the datatype of the existing property? If it displays a WYSIWYG based on the ontology, will it write HTML to the literal, but not change the datatype?

I've no problem with merging this as a basis for either doing an ontology change or annotation development before the next release. From the point of view of maintaining multi-lingual instances, I'm not in favour of an ontology change that would place restrictions on that/would require maintaining a local ontology that differs in this regard - but I don't know how much effort would be involved in using annotations to affect the choice of input form, as I've not tried to do that before.

awoods:
@mconlon17 : I believe your [comment](https://github.com/vivo-project/Vitro/pull/76#pullrequestreview-137892568) has been addressed. Shall we move this forward?

PR #77, opened 2018-07-06
Rel 1.10.0 rc pom.xml updates

roflinn:
Updated release profile, nexus-staging-maven-plugin plug-in version, repository, jenatools dependency

### DIFF
--- a/home/pom.xml
+++ b/home/pom.xml
@@ -69,13 +69,13 @@
         <dependency>
             <groupId>org.vivoweb</groupId>
             <artifactId>jena2tools</artifactId>
-            <version>1.2.0-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


home/pom.xml
```diff
@@ -69,13 +69,13 @@
         <dependency>
             <groupId>org.vivoweb</groupId>
             <artifactId>jena2tools</artifactId>
-            <version>1.2.0-SNAPSHOT</version>
+            <version>1.2.0</version>
```
awoods: I believe the release plugin requires the version to be at X.Y.Z-SNAPSHOT as a starting point, no?

awoods: :+1:

PR #79, opened 2018-07-12
[VIVO-1539] Moved stream reset inside loop so that every listener gets same state

grahamtriggs:
**[VIVO-1539](https://jira.duraspace.org/browse/VIVO-1539)**:

Moves the resetting of the serialized changes from outside the loop that notifies listeners, to inside, so that listeners will always get a stream that is ready to use, regardless of how many have been configured or the order.

# How should this be tested?
The only existing listener is the one that controls inferencing.
Primary testing should be to submit a change to the data, and ensure that any inferences are written correctly.

# Interested parties
@VIVO-project/vivo-committers


mconlon17:
Can we get a bit more on how this will be tested.  How do we see the bug? What will the behaviors be before (undesirable) and after (desirable).

PR #88, opened 2018-10-05
[VIVO-1605] Update to latest Error Prone plugin and configuration to support JDK 11

grahamtriggs:
**[VIVO-1605](https://jira.duraspace.org/browse/VIVO-1605)**: 

Updates the pom.xml to use the latest version of Google's Error Prone plugin and configuration, so that the build works with recently released JDKs

# What's new?
Nothing

# How should this be tested?
Do a full "mvn clean install" - the compilation should complete normally.

Should be tested with multiple JDK versions - should work for everything from JDK 8 through to the recently released JDK 11

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/installer/pom.xml
+++ b/installer/pom.xml
@@ -64,7 +64,7 @@
                 </plugin>
                 <plugin>
                     <artifactId>maven-compiler-plugin</artifactId>
-                    <version>3.5.1</version>
+                    <version>3.8.0</version>
                 </plugin>
                 <plugin>
                     <artifactId>maven-dependency-plugin</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -194,6 +194,27 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <!-- JDK 8 profile for the Error Prone plugin -->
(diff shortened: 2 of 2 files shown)


j2blake:
It seems inappropriate to use a SNAPSHOT version of a dependency. We could issue a VIVO release, and then a new SNAPSHOT of the dependency could be issued, potentially breaking our release.

Am I understanding this correctly?

grahamtriggs:
@j2blake That's correct - you shouldn't have SNAPSHOTs in releases for that reason.

For development, it's not necessarily a bad thing to have the SNAPSHOT to unblock an issue (e.g. incompatibility with the latest JDK), as long as you are committing to update the dependency before release.

As it is, I've updated the PR to the final release version, which is mentioned on GitHub, but it hasn't made it to the Maven central repositories yet - so it can't be merged until that happens.

roflinn:
Just to note: Tested Linux and Windows with RedHat OpenJDK 8 and 11 and Oracle OpenJDK 8 and 11. All Passed - Approved

PR #95, opened 2018-11-23
[VIVO-1543] Removing non existing flash.html file inclusion in page.ftl template

KitioFofack:
The PR removes non existing flash.html file inclusion in page.ftl template

**[JIRA Issue]: https://jira.duraspace.org/browse/VIVO-1543

It just removes an inclusion to non existent file

# How should this be tested?

* Reproduce the problem you are fixing (if applicable)
Switch to develop branch 
Rename the page.ftl template (to page.ftl.bak for instance) in your preferred theme. 
Run VIVO with that theme
You will get a 500 error on pages different from the home page 

Rename the page.ftl (to page.ftl.bak for instance) template in your preferred theme.
Run the application, it should work properly (no  500 error)

# Interested parties
@gneissone, @mconlon17, @awoods, @VIVO-project/vivo-committers

### DIFF
--- a/webapp/src/main/webapp/templates/freemarker/page/page.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/page/page.ftl
@@ -24,7 +24,7 @@
         <hr class="hidden" />
 
         <div id="contentwrap">
-  
+			<#include "flash.ftl">
             <div id="content">                      
                 ${body}
             </div> <!-- content -->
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/templates/freemarker/page/page.ftl
```diff
@@ -24,8 +24,7 @@
         <hr class="hidden" />
 
         <div id="contentwrap">
-            <#include "flash.html">
-
+
```
gneissone:
```suggestion
              <#include "flash.ftl">
```
Include the proper template instead of removing it.

PR #98, opened 2018-12-05
Vivo 1630 Year field keeps the same label even when language is changed 

alessandroGalasso:
**[JIRA Issue](https://jira.duraspace.org/projects/VIVO)**: https://jira.duraspace.org/browse/VIVO-1630

Use of localized Year from property file.
[Editors note] - 'Year' is hard-coded in english. It will appear as 'Year' regardless of the language selected.

# How should this be tested?
edit a person postion with a data interval
[Editors note] - This pull request creates an i18n value for 'Year.' Test the problem by selecting a different language than English and 'Year' will remain. After implementing the change (and the associated VIVO-languages pull request) 'Year' should change with the language. 

# Additional Notes:
i have VIVO-1649 as prerequsite to start my instance on blazegraph

# Interested parties
@VIVO-project/vivo-committers


webapp/src/main/webapp/i18n/all.properties
```diff
@@ -922,3 +922,7 @@ minimum_hour=Invalid entry. Please specify at least an Hour.
 year_month=Invalid entry. Please enter a Year and Month.
 decimal_only=Invalid entry.  A decimal point is allowed, but thousands-separators are not.
 whole_number=Invalid entry. Enter a whole number with no decimal point or thousands-separators.
+
+
+label.dateTimeWithPrecision.year_capitalized = Year
```
awoods: We should also see updates in the i18n files, no?
https://github.com/vivo-project/VIVO-languages/tree/master/webapp/src/main/webapp/i18n

awoods:
See: https://github.com/vivo-project/VIVO-languages/pull/19

PR #99, opened 2018-12-05
[VIVO-1656] Extension mechansim for passing properties to templates

grahamtriggs:
**[VIVO-1656](https://jira.duraspace.org/browse/VIVO-1656)**:

Adds an extension mechanism to the IndividualResponseBuilder that allows applications built on top of Vitro to register code to add extra parameters to be passed to the templates.

It also moves the Plum and AltMetrics extensions out of the Vitro code, as they are only applicable to VIVO. See the corresponding pull request in VIVO that uses the extension.

# What's new?
No change in functionality, this is an architectural improvement.

# How should this be tested?
Visit the page of any "individual" in Vitro. Should work exactly as it did before.

# Interested parties
@VIVO-project/vivo-committers


awoods:
Replaced by: https://github.com/vivo-project/Vitro/pull/125

PR #107, opened 2018-12-15
[VIVO-1667] Language filtering for CONSTRUCT queries

grahamtriggs:
**[VIVO-1667](https://jira.duraspace.org/browse/VIVO-1667)**:

Corrects a bug where one of the construct query methods in Language filtering RDF Service does not 

# What's new?
Nothing

# How should this be tested?
Requires multi-lingual features to be enabled.

Original problem can be observed by having multiple labels on a VCard URL (attached as hasURL to contact info of a profile). The websites field will display multiple entries, despite there only be a single VCard URL attached.

When applied, this patch will correct this, and only a single entry for the website will be displayed.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/rdfservice/filter/LanguageFilteringRDFService.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/rdfservice/filter/LanguageFilteringRDFService.java
@@ -22,6 +22,7 @@ import org.apache.jena.query.ResultSetFormatter;
 import org.apache.jena.rdf.model.Literal;
 import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.ModelChangedListener;
+import org.apache.jena.rdf.model.ModelFactory;
 import org.apache.jena.rdf.model.RDFNode;
 import org.apache.jena.rdf.model.Statement;
 import org.apache.jena.rdf.model.StmtIterator;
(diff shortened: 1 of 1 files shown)


awoods:
Good catch, @brianjlowe . Does that sound reasonable to you, @grahamtriggs ?

grahamtriggs:
Hmm.... this method signature was added to cut the unnecessary serialization / deserialization that was being done to take what is a Jena Model inside the RDFService to a Jena Model outside of the RDFService - something that was taking up a significant amount of CPU time without providing any function.

In almost all - if not all - cases, there is going to be an empty model passed into the method to receive the constructed triples, and if a model did happen to have triples already, they will likely have come through a (filtered) construct in the first place.

As a compromise, I've added a check to see if the Model is empty - if it is, there is no harm in using it directly. Then only in rare cases of the model already being populated will it incur the memory / CPU overhead of using a temporary model and adding the filtered statements to the destination.

brianjlowe:
Thanks, Graham.  This looks like a good approach.

As to the design considerations, given the expected use cases as stated, I struggle to understand the rationale for either the new interface method signature or the original implementation.  If it is expected that the supplied model will almost always be empty, would it not be better for the new RDFService method to be a function that returns a model object rather than a void method that mutates an existing object?  This would not only reduce the intellectual workload for the implementer (Is my code thread safe? Do I need to obtain a write lock on the model before adding to it?) but also eliminate the class of potential bugs resulting from unexpected side effects.

In the second scenario of adding to a model that has already been filtered, I would tend to imagine that the overhead involved in creating a temporary model is likely to be smaller than the overhead of repeatedly applying the filter function to a growing set of triples that have already been filtered.

mconlon17:
Looks the the first sentence under what does this pull request do? *Corrects a bug where one of the construct query methods in Language filtering RDF Service does not* is incomplete.  RDF Service does not **do what**?

The description of what the pull request does ("Nothing") is incorrect.  The pull request changes something.  What does it change?  some of this is in the "How to Test" section.  How to test can then be simplified -- create multiple language values and see they are all displayed before the pull request.  After the pull request, only one is displayed (which one?)

PR #109, opened 2019-01-04
Vivo 1642 dependency updates

jcreel:
** Upgrade dependencies in Vitro **
* * *

**[VIVO-1642](https://jira.duraspace.org/browse/VIVO-1642)**: 

Based on a pass through all the .pom files in Vitro and a comparison with what's available on https://mvnrepository.com, these changes bring all the dependency versions to the highest possible without breaking the build.  This is half of the work for card VIVO-1642, the other half being upgrading the dependencies in VIVO itself.  Another PR will be made to the develop branch of the VIVO repository when that work is complete.

# How should this be tested?
Ensure the build still works, tests pass, and the VIVO app deploys with the updated Vitro dependencies.

# Interested parties
@awoods

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -56,7 +56,7 @@
         <dependency>
             <groupId>de.mkammerer</groupId>
             <artifactId>argon2-jvm</artifactId>
-            <version>2.4</version>
+            <version>2.5</version>
         </dependency>
         <dependency>
             <groupId>org.vivoweb</groupId>
@@ -71,22 +71,22 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>javax.servlet</groupId>
+            <groupId>javax.servlet.jsp</groupId>
             <artifactId>jsp-api</artifactId>
-            <version>2.0</version>
+            <version>2.2</version>
             <scope>provided</scope>
         </dependency>
 
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.11</version>
+            <version>4.12</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.easymock</groupId>
             <artifactId>easymock</artifactId>
-            <version>3.5</version>
+            <version>4.0.2</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/api/src/test/resources/edu/cornell/mannlib/vitro/webapp/edit/n3editing/testEditConfig.json
+++ b/api/src/test/resources/edu/cornell/mannlib/vitro/webapp/edit/n3editing/testEditConfig.json
@@ -2,7 +2,7 @@
     "formUrl" : "${formUrl}",
     "editKey" : "${editKey}",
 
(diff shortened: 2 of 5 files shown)


api/pom.xml
```diff
@@ -71,22 +71,22 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>javax.servlet</groupId>
+            <groupId>javax.servlet.jsp</groupId>
             <artifactId>jsp-api</artifactId>
-            <version>2.0</version>
+            <version>2.2</version>
             <scope>provided</scope>
```
grahamtriggs: "Provided" APIs - such as JSP and Servlet spec should NOT be updated, unless we are making a conscious decision about the versions of Tomcat that we are supporting, otherwise there is a serious possibility that we could inadvertently break our support.

awoods: Thanks, @grahamtriggs . It looks like Tomcat 5.5 expects JSP 2.0:
http://tomcat.apache.org/whichversion.html

Tomcat 5.5 was EOL'd in 2012:
https://tomcat.apache.org/tomcat-55-eol.html

Tomcat 7 supports JSP 2.2. The latest release of Tomcat is 9. It seems safe to consciously require Tomcat 7 as a minimal requirement for VIVO. Thoughts?

gneissone: The 3.1 version of the servlet spec requires Tomcat 8 if I'm understanding http://tomcat.apache.org/whichversion.html correctly. Which I'm fine with as a requirement, though it's perhaps a bit controversial since Tomcat 7 EOL has not been announced. 

This PR updates the version from 2.5 to 3.1.0 in [installer/webapp/pom.xml](https://github.com/vivo-project/Vitro/blob/6e717446b4a1b3da0fcf0130f3d0cfd1ce8b75ed/installer/webapp/pom.xml#L159), but it looks like the dependency in [Vitro/api/pom.xml](https://github.com/vivo-project/Vitro/blob/f2656ba3d8ae67359567d5aecd92eb67ba743c62/api/pom.xml#L68) is already referencing 3.1.0? What's that mean?

awoods: @gneissone : agreed. Apparently we have already committed to Tomcat 7:
https://jira.duraspace.org/browse/VIVO-1375

Thoughts, @grahamtriggs ?

awoods: I second the idea of updating the VIVO requirements to Tomcat 8.x.
@grahamtriggs : Is there anything else that needs to be done in this PR to realize that?
..noting that the documentation will need to be updated:
https://wiki.duraspace.org/display/VIVODOC111x/System+Requirements#SystemRequirements-TomcatTomcat7orlater

grahamtriggs: If we are updating to Tomcat 8.x / Servlet 3.1.0 (and I'm happy to see this), then we should align JSP API to version 2.3 (as per http://tomcat.apache.org/whichversion.html).

Highly unlikely that this will require any changes in the code, but I won't vouch for that without doing it.

chenejac:
This is being closed due to inactivity; might be reopened in the future.

PR #116, opened 2019-04-11
[Vivo 1686] - Improve performance on the getPropertyList method

alessandroGalasso:
**[JIRA Issue](https://jira.duraspace.org/projects/VIVO)**: https://jira.duraspace.org/browse/VIVO-1686

It improves performance on the getPropertyList method of the BaseIndividualTemplateModel class.

# What's new?
the method is now parallel, allowing scalable parallelism on production server.

# How should this be tested?
- start the server
- log in as root
- open a person with a lot of properties.
- look at the server response time in the browser console

benchmark on Cineca server with the non parallel version:
sparqlContentTripleSource   -   getPropertyList processing time:  -60%
tdbContentTripleSource   -   getPropertyList processing time:  -50%

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/DataPropertyTemplateModelCallable.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/DataPropertyTemplateModelCallable.java
@@ -1,0 +1,39 @@
+package edu.cornell.mannlib.vitro.webapp.web.templatemodels.individual;
+
+import java.util.List;
+import java.util.concurrent.Callable;
+
+import edu.cornell.mannlib.vitro.webapp.beans.DataProperty;
+import edu.cornell.mannlib.vitro.webapp.beans.Individual;
+import edu.cornell.mannlib.vitro.webapp.controller.VitroRequest;
+
+public class DataPropertyTemplateModelCallable 
+implements Callable<DataPropertyTemplateModel> {
+
+	final DataProperty dp;
+	Individual subject;
+	final VitroRequest vreq;
+	final boolean editing;
+	final List<DataProperty> populatedDataPropertyList;
+	final String callableName;
+	
+	
+	public DataPropertyTemplateModelCallable(DataProperty dp, Individual subject, VitroRequest vreq,
+			boolean editing, List<DataProperty> populatedDataPropertyList,	String callableName) {
+		super();
+		this.dp = dp;
+		this.subject = subject;
+		this.vreq = vreq;
+		this.editing = editing;
+		this.populatedDataPropertyList = populatedDataPropertyList;
+		this.callableName = callableName;
+	}
+
+	@Override
+	public DataPropertyTemplateModel call() throws Exception {
+		DataPropertyTemplateModel dptm = new DataPropertyTemplateModel(dp, subject, vreq, editing, populatedDataPropertyList);
+		return dptm;
+		
+	}
+
+}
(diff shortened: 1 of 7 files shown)


gneissone:
Tested with SDB and everything appears to be functioning normally. I did notice a speed bump commensurate with your reports though I was not at all scientific about it. 
@grahamtriggs or @brianjlowe can you comment on the code itself?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/DataPropertyTemplateModelCallable.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class DataPropertyTemplateModelCallable 
+implements Callable<DataPropertyTemplateModel> {
+
+	final DataProperty dp;
+	Individual subject;
+	final VitroRequest vreq;
+	final boolean editing;
+	final List<DataProperty> populatedDataPropertyList;
+	final String callableName;
```
awoods: `callableName` never appears to be used. Can be removed?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/DataPropertyTemplateModelCallable.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.concurrent.Callable;
+
+import edu.cornell.mannlib.vitro.webapp.beans.DataProperty;
+import edu.cornell.mannlib.vitro.webapp.beans.Individual;
+import edu.cornell.mannlib.vitro.webapp.controller.VitroRequest;
+
+public class DataPropertyTemplateModelCallable 
+implements Callable<DataPropertyTemplateModel> {
+
+	final DataProperty dp;
```
awoods: These six fields can be `private`.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/GroupedPropertyList.java
```diff
@@ -136,12 +140,57 @@ is to make the get methods (getProperty and getPropertyAndRemoveFromList) method
         // Build the template data model from the groupList
         groups = new ArrayList<PropertyGroupTemplateModel>(
                 propertyGroupList.size());
+        
+        
+        if(!parallel) {
```
awoods: I believe this block will never be executed, because the one caller of this class now only passes in "parallel = true".

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/ObjectPropertyTemplateModelCallable.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class ObjectPropertyTemplateModelCallable 
+implements Callable<ObjectPropertyTemplateModel> 
+ {
+	final ObjectProperty op;
+	Individual subject;
+	final VitroRequest vreq;
+	final boolean editing;
+	final List<ObjectProperty> populatedObjectPropertyList;
+	final String callableName;
```
awoods: `callableName` never appears to be used. Can be removed?

alessandroGalasso:
@awoods  I will clean up the code of course; 
as for the compatibility issue, i think that we just have a lot of read-only select to a triplestores (inside a servlet), so they already need to be threadsafe from the triplestore point of view. 
If there is a problem there the triplestore is not very good for a production envirovment.
Now, of course, it is more likely for concurrent related problems to emerge in the webapp.
Let us consider, for example, the method getCustomListViewConfigFileName of class
ObjectPropertyDaoJena, it was responsable of a corrupt CustomListViewConfigFileMap
generation on parallel version without the synchronized keyword. 
In my opinion the performance improvement is too good to pass on this one, half time and we are
also fetching statements data.

alessandroGalasso:
new PR https://github.com/vivo-project/Vitro/pull/120

PR #118, opened 2019-04-26
User Interface for DataDistributor API

grahamtriggs:
**[VIVO-1691](https://jira.duraspace.org/browse/VIVO-1691)**:

# What's new?
Adds a user interface for administering the configuration of the DataDistributor API.

A link is added to the Site Admin menu, which takes you to a screen with a list of DataDistributor and GraphBuilders.

New entries for either can be created, and existing entries can be edited (and deleted from the edit page). Note that anything that is loaded from an "everytime" file in the RDF directory is displayed as read only and can not be edited or deleted.

Only Site Admins are given access, but this can be changed from the authorisation configuration.

# How should this be tested?
Log in as a site admin / root. From the site admin menu, select the entry to "manage datadistributors".

Add new distributors and graphbuilders of the different types.
Enter various configurations.
Edit and delete configurations.

Access saved distributors via the DataDistributor api (/api/dataRequest/<actionName>)

# Additional Notes:
This pull request contains the code of the DataDistributor API (Cornell's repository is now marked as "archived").

There are two minor modifications to Cornell's code:

1) The actionName in the AbstractDataDistributor has been marked as minOccurs / maxOccurs = 1 (so that it shows up as a required field).

2) An AbstractGraphBuilder has been introduced, and includes a builderName (similar to actionName). This is not a functional change in how the DataDistributor API works, but a convenience so that when selecting a graph builder as a provider to a data distributor, you can see a name rather than a URI.

In all other ways, the DataDistributor API is unchanged.

# Interested parties
@VIVO-project/vivo-committers


awoods:
Instead of pulling the https://github.com/cul-it/vivo-data-distribution-api project into Vitro, would you be open to pulling the project into the https://github.com/vivo-project GitHub org?
Breaking the code into modules, and reducing the size of the core codebase is one of the technical objectives coming from the https://wiki.duraspace.org/display/VIVO/Product+Direction+for+2019 .

grahamtriggs:
@awoods There are practical problems with that. This patch is adding a user interface - so it is adding templates, changing existing templates to add links, and adding entries into the i18n files.

In turn, those templates have a dependency on back-end classes that are being added here to save te configuration, both of these have dependencies on the data-distribution-api, and all of these have dependencies on the Vitro API (which also has been modified to add an authorization policy)

If we want to ship a UI for managing the data distribution config (and/or an application that has the data distribution api enabled by default), then this code needs to be compiled before the webapp - but with (or after) the vitro api.

gneissone:

chenejac:
This is being closed due to inactivity; might be reopened in the future.

PR #120, opened 2019-05-06
[Vivo 1686] -  Improve performance on the getPropertyList method

alessandroGalasso:
**[JIRA Issue](https://jira.duraspace.org/projects/VIVO)**: https://jira.duraspace.org/browse/VIVO-1686

It improves performance on the getPropertyList method of the BaseIndividualTemplateModel class.

# What's new?
the method is now parallel, allowing scalable parallelism on production server.

# How should this be tested?
- start the server
- log in as root
- open a person with a lot of properties.
- look at the server response time in the browser console

benchmark on Cineca server with the non parallel version:
sparqlContentTripleSource   -   getPropertyList processing time:  -60%
tdbContentTripleSource   -   getPropertyList processing time:  -50%

# Interested parties
@VIVO-project/vivo-committers


api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/BaseIndividualTemplateModel.java
```diff
@@ -104,7 +104,7 @@ public String getRdfUrl() {
         }
     }
 
-    public GroupedPropertyList getPropertyList() {
+    public GroupedPropertyList getPropertyList() throws Exception {
```
grahamtriggs: Not a good idea to add a generic exception here. Should at least be limited to throwing the exceptions that occur in the method, but also consider whether they can be handled within the method and not changing the signature.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/DataPropertyTemplateModelCallable.java
```diff
@@ -0,0 +1,33 @@
+package edu.cornell.mannlib.vitro.webapp.web.templatemodels.individual;
```
grahamtriggs: Missing header comment - can simply copy from other existing files.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/DataPropertyTemplateModelCallable.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.List;
+import java.util.concurrent.Callable;
+
+import edu.cornell.mannlib.vitro.webapp.beans.DataProperty;
+import edu.cornell.mannlib.vitro.webapp.beans.Individual;
+import edu.cornell.mannlib.vitro.webapp.controller.VitroRequest;
+
+public class DataPropertyTemplateModelCallable implements Callable<DataPropertyTemplateModel> {
```
grahamtriggs: Should we make this Callable<PropertyTemplateModel>, so that the code that process the results can use Future<PropertyTemplateModel> rather than Future<?> ?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/GroupedPropertyList.java
```diff
@@ -67,7 +71,7 @@ is to make the get methods (getProperty and getPropertyAndRemoveFromList) method
     private List<PropertyGroupTemplateModel> groups;
 
     GroupedPropertyList(Individual subject, VitroRequest vreq,
-            boolean editing) {
+            boolean editing) throws Exception {
```
grahamtriggs: Not a good idea to add a generic exception here. Should at least be limited to throwing the exceptions that occur in the method, but also consider whether they can be handled within the method and not changing the signature.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/GroupedPropertyList.java
```diff
@@ -136,10 +140,36 @@ is to make the get methods (getProperty and getPropertyAndRemoveFromList) method
...
-            groups.add(new PropertyGroupTemplateModel(vreq, propertyGroup,
-                    subject, editing, populatedDataPropertyList,
-                    populatedObjectPropertyList));
+            groups.add(new PropertyGroupTemplateModel(vreq, propertyGroup, subject, editing,
+                    populatedDataPropertyList, populatedObjectPropertyList, executor));
+        }
+        
+        // wait all
+        for (PropertyGroupTemplateModel pgtm : groups) {
+            List<Future<?>> futures = pgtm.getFutures();
```
grahamtriggs: Would be better if this was the more precise List<Future<PropertyTemplateModel>>, but requires the Callable models to be the Callable<PropertyTemplateModel>.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/ObjectPropertyTemplateModelCallable.java
```diff
@@ -0,0 +1,41 @@
+package edu.cornell.mannlib.vitro.webapp.web.templatemodels.individual;
```
grahamtriggs: Missing header comment - can simply copy from other existing files.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/PropertyGroupTemplateModel.java
```diff
@@ -33,14 +35,20 @@
     private final String name;
     private final List<PropertyTemplateModel> properties;
 
-
+    private List<Future<?>> futures = null;
```
grahamtriggs: Would prefer Future<PropertyTemplateModel> - see other comments.

awoods:
Thanks for the review, @grahamtriggs .

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/jena/ObjectPropertyDaoJena.java
```diff
@@ -1087,7 +1087,7 @@ protected void processQuerySolution(QuerySolution qs) {
     Map<FullPropertyKey, String> customListViewConfigFileMap;
 
     @Override
-    public String getCustomListViewConfigFileName(ObjectProperty op) {
+    public synchronized String getCustomListViewConfigFileName(ObjectProperty op) {
```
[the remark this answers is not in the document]
alessandroGalasso: i can annotate the interfaces with @ThreadSafe and then if somenone want to improve the code and remove the "synchronized" from the implementation has to keep that in mind. I may say that the bottleneck on the API
is one or more  orders of magnitude lower than improvment on the overall processing time.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/GroupedPropertyList.java
alessandroGalasso: maybe we can make 

ObjectPropertyTemplateModelCallable implements Callable<PropertyTemplateModel>  
DataPropertyTemplateModelCallable implements Callable<PropertyTemplateModel>

then   

List<Future<PropertyTemplateModel>> futures = pgtm.getFutures();

grahamtriggs: Ah - yes. That's better! (It's also what I typed into the review, but the editor must have taken it to be html and mangled it).

PropertyTemplateModel as the generic class throughout, rather than an indeterminate Future type.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/GroupedPropertyList.java
```diff
@@ -136,10 +140,36 @@ is to make the get methods (getProperty and getPropertyAndRemoveFromList) method
...
+                            properties.add(tm);
+                        }
+                    } else if (f.get() instanceof DataPropertyTemplateModel) {
+                        DataPropertyTemplateModel dptm = (DataPropertyTemplateModel) f.get();
+                        properties.add(dptm);
+                    } else {
+                        log.error("unexpected class " + f.get().getClass());
+                    }
+                } catch (InterruptedException | ExecutionException e) {
+                    throw e;
```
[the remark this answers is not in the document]
grahamtriggs: I would probably go IllegalStateException(e) - but yes, a derivation of RuntimeException seems preferable here to generic checked exceptions.

Also, I noticed that the try { } finally { } is within the for loop processing futures, when it should wrap everything from immediately after the executor is created to the end of the two for loops that follow it (up to the if (!editing) { pruneEmptyProperties(); }.

Given the throwing of a RuntimeException, the catch block can also be relocated with the finally at no real detriment.

grahamtriggs:
> Fortunately, looking at the system monitor the load of the Java application / mysql seems to stay quite low with the parallel code (Chrome is the main resource hog in rendering the html). And simultaneous access also appear to be OK from what I've been able to test so far.

Having done more load testing, I'm going to have to revise this observation.

For the dataset that I'm using locally - on SDB - I have a profile that loads in ~4.5 seconds in the non-parallel code.

For a single user, this drops to ~3.5 seconds.

However, on ramping up the number of users things change quite quickly.

Single-threaded code:
Maxes out CPU around 8 concurrent users, at 20 users the request time drops to ~13.5 seconds.

Parallel code:
Maxes out CPU around 4 concurrent users, at 20 users the request time drops to ~20 seconds.

This looks related to running on a hyper-threaded quad-core, and the overhead of thread switching once that is exhausted.

@alessandroGalasso I would be interested in more details about your setup - are you running BlazeGraph on a separate server? Do you know what CPU resources the server(s) have?

Can you comment on how request time is affected by having 20 simultaneous users in your setup?

I believe you mentioned somewhere the overhead of creating connections for the SPARQL queries - is there something that can be done here (HTTP/1.1, connection pooling, etc.) that could be done to reduce this overhead?

alessandroGalasso:
@grahamtriggs i have done some testing.

Test with root authenticated admin users  ( concurrent: 1,5,10 ) and a FacultyMember editable page  


Test Server
8vcpu 8GB Blaze / 6GB VIVO / 4GB solr ,  virtual machine 

Jmeter Configuration

HTTP Request Default   		http  servename  8080

HTTP Cookie Manager

Once Only Controller
 HTTP Request login		
   /vivo-webapp(TDB/BLAZE/parallel/series)/authenticate  loginName:xx	loginPassword:yy loginForm:Log+in

HTTP Request authPerson    	/vivo-webapp(TDB/BLAZE/parallel/series)/display/nYYYY

1) 
Tests Blaze-parallel  Ramp-Up: 1s Loop : 10  ; concurrent Users root: 1,5,10	

Sam	Min	 	Aver  	Max
10	3544	4054	4542	
50	6323	7623	9132	
100	9848	16172	22842 (overall time 2m49s)

2)
Tests Blaze-series  Ramp-Up: 1s Loop : 10  ; concurrent Users root: 1,5,10	

Sam	Min	 	Aver  	Max

10	7867	8373	8829	
50	8862	9673	11196	
100	13167	15446	17972	(overall time 2m39s)

3)
Tests TDB-parallel  Ramp-Up: 1s Loop : 10  ; concurrent Users root: 1,5,10

Sam	Min	 	Aver  	Max   
10	3288	3826	4376	
50	5739	7122	8945	
100	9288	13146	17779 (overall time 2m:17s)


4)
Tests TDB-series  Ramp-Up: 1s Loop : 10  ; concurrent Users root: 1,5,10

Sam	Min	 	Aver  	Max   
10	6503	7184	7575
50	7570	8540	9525
100	6708	14043	18977 (overall time 2m:36s)


first of all i think this test is to hard for the server , and it doesnt keep up. 

Then as we can see that the average response time for parallel and series version
as we put more concurrent users go up a lot because 
the load of this test is really excessive for the server,
and of course more thread means more work and maybe less FIFO approach.

But i have to get the server i need to have an acceptable response time,
and if i look to response times of more than 10s to see 
who is better in that range maybe i'm already out of scope.

I have to do the benchmarks at least for response time in (0,10) sec interval 
and get the server i need to stay there.

grahamtriggs:
@alessandroGalasso Can I ask why you are testing with a root user?

Yes, the performance when using a root user can be poor, and this is specifically because it displays the unbounded properties for what are used in faux properties:

i.e. Authors are faux properties based on Related By (with an Authorship context). The root user displays both Authors, but also everything that is related by (including the authorship relationships)

So wherever there exists a faux property - a person with thousands of publications linked to them - being logged in as a root user is asking the application to do roughly twice the amount of work. To display the publications as "selected publications", and as authorship nodes in related by.

This is only an issue when using the root user - "related by" is not configured to be displayed as a regular property for any of the defined roles in the system.

In other words, simply logging in as a site admin - or accessing a page not logged in (as most users of a public app would do) - will show similar kind of response times as using parallel code logged in as root user.

It is not recommended to be logging in a root user unless it is absolutely necessary. Yes, it's convenient from a developer's point of view, and on rare occasions it might be required to access a function or property that has been locked down from every other role in the system. But in real world use, nobody should ever be logging in as root and accessing the content pages as a matter of course.

What it comes down to is do you need to have access to the "related by" list of properties in "view all" (which are otherwise being shown as publications, as authors, etc.)? Because if you don't (and usually you don't, because they are displayed elsewhere), then you are better off logging in as a user with a defined role, or making changes to the authorization code so that the "bare" faux properties can be hidden from root users, as it doesn't just exhibit the same savings in response time - it is halving the amount of work the server is doing.

Yes, there are some - more modest - benefits in response time for non-root access with the parallel code, on a lightly accessed system. But the more I look at it, the more I feel those benefits are outweighed by the additional complexity, increased resource use, possibility of new errors, increased difficulty in debugging and profiling the code - for anything other than the root user access case.

galasso:
@grahamtriggs very good point , I'll need a "more real world" test plan, self-editor users should be a good choice and a PoolingHttpClientConnectionManager  could be of help too.

alessandroGalasso:
@grahamtriggs  @awoods   
first of all thank you for your detailed analysis of my PR.

Today I tested on my linux (test) server with self-editor users (clicking non stop).
I think that with low load a self-editor has an almost accettable response time (2,2s) 

Also i think that "getPropertyList method" give the CPU a lot of load and maybe some improvement  can be done there.

Anyway we need to pass the pre-production testiong phase and
from the university test group point of view (who have to approve/buy it)  
the following response times are relevant :
- the response time to open a non cached resource for the public user
- the response time for the "logged in user" to open a page 
And we get a better chance of success on a demo server with no load and parallel code

Then we can go production and i hope to get a lesser than 2 sec response time on low load;
i also hope that HW tech improvments will help fix slow vivo response times with parallel code
as the HW get better performance.

Test results my server:

Tests Blaze parallel  Ramp-Up: 1s Loop : 10  ; concurrent: 1,3,5,10 logged in self editors	

Samples	Min	 	Aver  	Max   	load
10		2104	2299	2768 	CPU 20%
30		2824	3245	4456	CPU 80%
50		3724	4651	6105 	CPU 100%
100		7273	9385	12349	CPU 100%

Tests Blaze series  Ramp-Up: 1s Loop : 10  ; concurrent: 1,3,5,10	 logged in self editors	

Samples	Min	 	Aver  	Max   	load
10		4402	4600	5221	CPU 12%
30		4540	4978	5590	CPU 32%
50		4807	5413	6505	CPU 50%
100		7092	9326	12307	CPU 100%


So this PR cames with pros and cons, i like and will keep it it for my web app instances, but your decision to make for community code base.

grahamtriggs:
@alessandroGalasso FYI, I have recently encountered an issue with self editors, in terms of how long it can take to process the authortizations for connected resources (not the identify of the person itself).

This is due to the way relationships are checked, and may be having an impact on when you are testing with a self editor - it would be useful if you tested with e.g. site admin - to see if there is a difference in your cases between that and self editor.

I will be taking a look at how we can improve the relationship check, but I have some vacation booked.

My recommendation is to hold off merging a parallel code until I've had a chance to evaluate that, so that we can make a real judgement about the trade-offs involved, but in any case thanks for making this available.

alessandroGalasso:
@grahamtriggs  , no rush, enjoy holidays; I'll leave some more stats on different roles for my system:

1,3 concurrent logged in users:

self editor
Samples	Min	 	Aver  	Max   	
10		2104	2299	2768
30		2937	3491	4733	

editor
Samples	Min	 	Aver  	Max   
10		3345	3773	4180	
30		4337	4831	5456	

curator
Samples	Min	 	Aver  	Max   
10	3835	4244	4906	
30	4769	5587	6863	

site admin
Samples	Min	 	Aver  	Max   
10		3402	3911	4599	
30		4609	5735	6913

nothing bat to be in a cpu sweet spot.

alessandroGalasso:
better concurrent performance with  [27fed26](https://github.com/vivo-project/Vitro/pull/120/commits/27fed26dfff470a9a986f8301ef48ba339dc41f2)

awoods:
@alessandroGalasso : Do you have performance numbers with your latest update?

alessandroGalasso:
@awoods   with last update have better performance on heavy and medium load.
coudl be interesting to test this last version on other instances.

Tests Blaze parallel Ramp-Up: 1s Loop : 10 ; concurrent: 1,3,5,10,10 logged in self editors
singleton pattern executor

Samples Min Aver Max load
10	2148	2246	2629	
30	2484	2887	3380	
50	3236	3824	5206	
100	4381	7109	9393	
100	4533	6740	10192

api/src/main/java/edu/cornell/mannlib/vitro/webapp/web/templatemodels/individual/ExecutorFactory.java
```diff
@@ -0,0 +1,15 @@
+/* $This file is distributed under the terms of the license in LICENSE$ */
+
+package edu.cornell.mannlib.vitro.webapp.web.templatemodels.individual;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+public final class ExecutorFactory {
+	
+	private static final ExecutorService executor = Executors.newCachedThreadPool();
```
grahamtriggs: This will cause the application to not be unloaded and leak resources.

Threads created within the lifetime of the application need to be terminated within the lifetime of the application. If you are having a shared executor that lasts over the lifetime of the application, then you need to hook in a servlet context listener and ensure that the the executor is cleaned up when the context is destroyed.

alessandroGalasso: true

awoods:
Where does this pull-request stand? @alessandroGalasso ? @grahamtriggs ?

gneissone:

PR #131, opened 2020-01-20
[VIVO-1737] Update Error Prone

grahamtriggs:
**[VIVO-1737](https://jira.lyrasis.org/browse/VIVO-1737)**:

Updates the Error Prone plugin to for compile time analysis to the latest release, and moves the configuration to a profile that is enabled for JDK 1.8 - 11.

Builds with Error Prone fail when using JDK 12+, so by not enabling on those JDKs enables people to use them for their environment - Vitro/VIVO builds on JDK 13, but it is not tested.

This also allows developers to disable the Error Prone plugin from the command line, however I would not recommend this in general as it quickly and easily catches bad code.

# What's new?
Nothing specific, although the Error Prone plugin is now faster and includes some extra checks

# How should this be tested?
Run mvn clean install - build should complete.
Ideally, test with multiple JDK versions

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/IndividualImpl.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/IndividualImpl.java
@@ -20,6 +20,7 @@ import edu.cornell.mannlib.vitro.webapp.filestorage.model.ImageInfo;
 /**
  * Represents a single entity record.
 */
+@SuppressWarnings("ComparableType")
 public class IndividualImpl extends BaseResourceBean implements Individual, Comparable<Individual> {
 	/**
 	 * This can be used as a "not initialized" indicator for a property that
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/grefine/JSONReconcileServlet.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/grefine/JSONReconcileServlet.java
@@ -469,7 +469,7 @@ public class JSONReconcileServlet extends VitroHttpServlet {
         return queryStr.replaceAll("\\s+", "\\\\ ");
     }
 
-    public class SearchResult implements Comparable<Object> {
+    public class SearchResult implements Comparable<SearchResult> {
         private String label;
         private String uri;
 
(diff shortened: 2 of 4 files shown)


gneissone:
No issues building with the changes. I am not seeing any warnings about Comparable<> when building on the develop branch, do you? I am running 1.8 [Java(TM) SE Runtime Environment (build 1.8.0_152-b16)] so maybe it only happens with more modern versions of Java.

grahamtriggs:
No, the code analysis is consistent on all JDKs. You are not seeing a warning (or rather, the build failing), because the pull request also includes the necessary changes to pass the checks - in two cases setting the scope of Comparable to the class it is being implemented on (because it is only being used in lists of that class), and in the other suppressing the warning as it is scoped to an interface not a specific implementation of it.

gneissone:
I mean I'm not seeing the warnings when building without your changes, I'm not sure how to replicate that part of the existing behavior mentioned in the [Jira issue](https://jira.lyrasis.org/browse/VIVO-1737).

grahamtriggs:
Ah, you aren't supposed to replicate that part of the ticket - I was noting that Error Prone now identifies these issue and that this patch includes fixes to them (otherwise the build would be broken).

If you want to replicate them, you could reverse the changes to IndividualImpl / JsonReconcileServlet / AutocompleteServlet that are part of this patch.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/IndividualImpl.java
```diff
@@ -20,6 +20,7 @@
 /**
  * Represents a single entity record.
 */
+@SuppressWarnings("ComparableType")
```
gneissone: @grahamtriggs is this suppressed because it is a false positive (per errorprone [documentation](https://errorprone.info/bugpattern/ComparableType)) or because it's not a simple fix to pass the errorprone check?

awoods:
Error Prone runs and VIVO builds as expected on Java 11.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/IndividualImpl.java
grahamtriggs: I'm treating it as a false positive, because it is using an interface that the IndividualImpl implements. Therefore, you would expect that when the comparable interface is being used, it is used to compare lists containing Individual, which could be any implementation.

It's reasonable to expect that we have lists of Individual (the interface), and that we would want to compare them, regardless of the underlying implementation.

For the two SearchResult occurrences, they are only used in Lists containing only the concrete class, and in any case, suggesting that they can be compared to any Object is too vague.

PR #137, opened 2020-03-20
[VIVO-1755] - Better error handling when reasoner disabled

gneissone:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1755)**: VIVO-1755

Fixes an issue that makes the site admin page inaccessible and the editor for data properties to break when the reasoner is disabled.

# What's new?
- Instead of throwing an error and stack trace up on the front end, the error is sent to the log. 
- An info box is displayed on the site admin page noting that the reasoner is disabled. The info box is already displayed when the reasoner status is abnormal, so I reused the functionality. 

# How should this be tested?
Reproduce the problem by commenting out these lines in tomcat/webapps/vivo/WEB-INF/resources/startup_listeners.txt

```
# edu.cornell.mannlib.vitro.webapp.application.ApplicationImpl$ReasonersSetup
# edu.cornell.mannlib.vitro.webapp.servlet.setup.SimpleReasonerSetup
```
Go to localhost:8080/vivo/siteAdmin.
You should get an error.

Screen Shot 2020-03-20 at 12 45 51 PM

Rebuild with changes. Site admin panel should work once again. Error should be printed to vivo.all.log, and stack trace should be printed to log if the log level is set to debug. 

Screen Shot 2020-03-20 at 12 38 46 PM

Reproduce the data property editor error by turning reasoning off and trying to add a new record for a common data property, e.g. overview. An error will be thrown: java.lang.IllegalStateException: JFactTBoxReasonerModule has not been started. Rebuild with changes and you should be able to add the data property.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
@@ -209,7 +209,7 @@ public class BaseSiteAdminController extends FreemarkerHttpServlet {
                 }
             } catch (IllegalStateException e) {
                 error = "The inferencing engine is disabled.";
-                log.debug("Status of reasoner could not be determined.", e);
+                log.debug("Status of reasoner could not be determined. It is likely disabled.", e);
             }
 
             if (error != null) {
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
@@ -199,12 +199,18 @@ public class BaseSiteAdminController extends FreemarkerHttpServlet {
 
             String error = null;
             String explanation = null;
-            TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
-            if (!status.isConsistent()) {
-                error = "INCONSISTENT ONTOLOGY: reasoning halted.";
-                explanation = status.getExplanation();
-            } else if ( status.isInErrorState() ) {
-                error = "An error occurred during reasoning. Reasoning has been halted. See error log for details.";
+            try {
+                TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
+                if (!status.isConsistent()) {
+                    error = "INCONSISTENT ONTOLOGY: reasoning halted.";
+                    explanation = status.getExplanation();
+                } else if ( status.isInErrorState() ) {
+                    error = "An error occurred during reasoning. Reasoning has been halted. See error log for details.";
+                }
+            } catch (IllegalStateException e) {
+                error = "The inferencing engine is disabled.";
+                log.warn("Status of reasoner could not be determined.");
+                log.debug(e);
             }
 
             if (error != null) {
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
@@ -208,7 +208,7 @@ public class BaseSiteAdminController extends FreemarkerHttpServlet {
                     error = "An error occurred during reasoning. Reasoning has been halted. See error log for details.";
                 }
             } catch (IllegalStateException e) {
-                error = "The inferencing engine is disabled.";
+                error = "The inferencing engine is disabled. Data entered manually may exhibit unexpected behavior prior to inferencing.";
                 log.debug("Status of reasoner could not be determined. It is likely disabled.", e);
             }
 
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
@@ -209,8 +209,7 @@ public class BaseSiteAdminController extends FreemarkerHttpServlet {
                 }
             } catch (IllegalStateException e) {
                 error = "The inferencing engine is disabled.";
-                log.warn("Status of reasoner could not be determined.");
-                log.debug(e);
+                log.debug("Status of reasoner could not be determined.", e);
             }
 
             if (error != null) {
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/jena/JenaAdminActions.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/jena/JenaAdminActions.java
@@ -217,7 +217,7 @@ public class JenaAdminActions extends BaseEditController {
 			}
 		}
 		catch (IllegalStateException e) {
-			log.debug("Status of reasoner could not be determined. It is likely disabled", e);
+			log.warn("Status of reasoner could not be determined. It is likely disabled", e);
 		}
     }
 
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/jena/JenaAdminActions.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/jena/JenaAdminActions.java
@@ -192,29 +192,33 @@ public class JenaAdminActions extends BaseEditController {
 	}
 
     private void printRestrictions() {
-    	TBoxReasonerModule reasoner = ApplicationUtils.instance().getTBoxReasonerModule();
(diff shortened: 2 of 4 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
```diff
@@ -199,12 +199,17 @@ protected ResponseValues processRequest(VitroRequest vreq) {
...
+                TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
+                if (!status.isConsistent()) {
+                    error = "INCONSISTENT ONTOLOGY: reasoning halted.";
+                    explanation = status.getExplanation();
+                } else if ( status.isInErrorState() ) {
+                    error = "An error occurred during reasoning. Reasoning has been halted. See error log for details.";
+                }
+            } catch (IllegalStateException e) {
+                error = "The inferencing engine is disabled.";
+                log.debug("Status of reasoner could not be determined.", e);
```
awoods: It seems like the debug message would be something more like the `error`:
```
Status of reasoner could not be determined because the inferencing engine is disabled.
```

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/fields/IndividualsViaVClassOptions.java
```diff
@@ -102,8 +107,13 @@ public FieldOptions setDefaultOptionLabel(String label){
     }
 
     protected boolean isReasoningAvailable(){
-    	TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
-    	return status.isConsistent() && !status.isInErrorState();
+    	try {
+            TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
+            return status.isConsistent() && !status.isInErrorState();
+            } catch (IllegalStateException e) {
```
awoods: Minor: Formatting is a bit off.
```
} catch (...
```
..should be lined up with `try`.

gneissone:
> It appears that: `ApplicationUtils.instance().getTBoxReasonerModule()` is also being called at:
> https://github.com/vivo-project/Vitro/blob/master/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/jena/JenaAdminActions.java#L195
> 
> ..it may be worth include a try/catch there as well.

Addressed your comments. I added the logic to JenaAdminActions.java but I would flag that servlet for possible removal, I've never seen it documented or used. I did test the various actions in the servlet and they appear to work, FWIW. I had to adjust the logging level from trace to info for the printRestrictions action to see the output since I wasn't about to set the entire application log to trace.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
```diff
@@ -199,12 +199,17 @@ protected ResponseValues processRequest(VitroRequest vreq) {
...
+                TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
+                if (!status.isConsistent()) {
+                    error = "INCONSISTENT ONTOLOGY: reasoning halted.";
+                    explanation = status.getExplanation();
+                } else if ( status.isInErrorState() ) {
+                    error = "An error occurred during reasoning. Reasoning has been halted. See error log for details.";
+                }
+            } catch (IllegalStateException e) {
+                error = "The inferencing engine is disabled.";
+                log.debug("Status of reasoner could not be determined. It is likely disabled.", e);
```
awoods: Suggest changing the logging level to `warn`.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/jena/JenaAdminActions.java
```diff
@@ -192,29 +192,33 @@ private String testWriteXML() {
...
+					log.trace("Has value: ");
+				}
+				log.trace("On property "+rest.getOnProperty().getURI());
+				for (Resource inst : rest.listInstances().toList() ) {
+					log.trace("     "+inst.getURI());
+				}
+			}
+		}
+		catch (IllegalStateException e) {
+			log.debug("Status of reasoner could not be determined. It is likely disabled", e);
```
awoods: Suggest changing the logging level to `warn`.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/jena/DataPropertyDaoJena.java
```diff
@@ -360,8 +360,13 @@ public List getAllExternalIdDataProperties() {
     }
 
     protected boolean reasoningAvailable() {
-    	TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
-    	return status.isConsistent() && !status.isInErrorState();
+        try {
+            TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
+            return status.isConsistent() && !status.isInErrorState();
+        } catch (IllegalStateException e) {
+            log.debug("Status of reasoner could not be determined. It is likely disabled.", e);
```
awoods: Suggest changing the logging level to `warn`.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/fields/IndividualsViaVClassOptions.java
```diff
@@ -102,8 +107,13 @@ public FieldOptions setDefaultOptionLabel(String label){
     }
 
     protected boolean isReasoningAvailable(){
-    	TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
-    	return status.isConsistent() && !status.isInErrorState();
+    	try {
+            TBoxReasonerStatus status = ApplicationUtils.instance().getTBoxReasonerModule().getStatus();
+            return status.isConsistent() && !status.isInErrorState();
+        } catch (IllegalStateException e) {
+            log.debug("Status of reasoner could not be determined. It is likely disabled.", e);
```
awoods: Suggest changing the logging level to `warn`.

awoods:
Although I do not get the stacktrace once I deploy with this patch, I do not see the notification in the siteAdmin??

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
gneissone: I don't think this warrants a warning if the reasoner was intentionally disabled. I wouldn't want my log littered with stack traces if the application was running as I expected. If anything, perhaps a warning could be thrown during the startup tests if the reasoner isn't running if deemed necessary. Thoughts?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/jena/DataPropertyDaoJena.java
gneissone: Same as above, I don't think this warrants a warn-level log because if the reasoner is unavailable it was likely intentional and the application will run as normal without it. This method is intended to determine if reasoning is available and returns false if not, that's not an error to me.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/jena/JenaAdminActions.java
gneissone: Ok, I agree this one deserves a warning since the function doesn't work at all if the reasoner is disabled.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/BaseSiteAdminController.java
awoods: Flagging in the startup-status would be great. Not changing the logging level here is fine.

gneissone: Looked a little into this. It would be easy to copy the existing TBox reasoner check into a smoke test. Not sure if anything similar exists for the ABox (Simple) reasoner. The ABox reasoner doesn't appear to be abstracted out into an interface, so if I'm understanding things correctly we'd have to test for the Simple Reasoner explicitly, which could theoretically make it harder to swap out reasoners in the future. I believe this was done once in the past (Pellet -> Simple).

gneissone: Per discussion on developer call today, will not add to startup smoke tests.

PR #138, opened 2020-03-22
[VIVO-1756] - Make data property richtext editor option selectable from UI

gneissone:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1756)**: [VIVO-1756]

Follows up on comments in https://github.com/vivo-project/Vitro/pull/76

Makes the rich text editor for data properties selectable via the UI

# What's new?
Adds a drop down menu to the data property settings form. Selecting 'rich text' from this menu results in a triple being added for the data property in the form <data prop uri> <http://vitro.mannlib.cornell.edu/ns/vitro/0.7#editing> "HTML" .

# How should this be tested?
Add a new data property or edit an existing one.

Screen Shot 2020-03-21 at 7 22 03 PM

Add the bottom of the form, there should be a new option for enabling the rich text editor.

Screen Shot 2020-03-21 at 7 22 16 PM

Select rich text and confirm that the value 'HTML' is populated 

Screen Shot 2020-03-21 at 7 22 29 PM

Add or edit an instance of the property, confirm it uses the rich text editor.

Screen Shot 2020-03-21 at 7 22 39 PM

Edit the property settings again to switch the editor back to the default. Confirm the rich text editor is no longer used.

Screen Shot 2020-03-21 at 7 32 23 PM

# Additional Notes:
'HTML' is currently the only value that has any effect, and it is hard coded into both the default form (https://github.com/vivo-project/Vitro/blob/master/webapp/src/main/webapp/templates/freemarker/edit/forms/defaultDataPropertyForm.ftl#L34) and the proposed DatapropRetryController.java here. Perhaps there's a more eloquent way to handle this but given the current situation it seemed sufficient. 

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/edit/DatapropEditController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/edit/DatapropEditController.java
@@ -42,7 +42,7 @@ public class DatapropEditController extends BaseEditController {
 
     	VitroRequest vreq = new VitroRequest(request);
 
-        final int NUM_COLS=18;
+        final int NUM_COLS=19;
 
         String datapropURI = request.getParameter("uri");
 
@@ -73,8 +73,9 @@ public class DatapropEditController extends BaseEditController {
         results.add("display tier");          // column 14
         results.add("display limit");         // column 15
         results.add("custom entry form");     // column 16
-        results.add("URI");                   // column 17
-        results.add("publish level");         // column 18
+        results.add("editing");               // column 17
+        results.add("URI");                   // column 18
+        results.add("publish level");         // column 19
 
         results.add(dp.getPickListName()); // column 1
         results.add(dp.getPublicName() == null ? "(no public label)" : dp.getPublicName()); // column 2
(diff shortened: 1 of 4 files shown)


awoods:
@gneissone : Is there value in additionally enabling a "textarea" option for larger textual content, per https://github.com/vivo-project/Vitro/pull/76#pullrequestreview-137568614 ?

gneissone:
> @gneissone : Is there value in additionally enabling a "textarea" option for larger textual content, per [#76 (review)](https://github.com/vivo-project/Vitro/pull/76#pullrequestreview-137568614) ?

Yes, I think so. Do you have a suggestion for what value of the editing property should result in a bigger plaintext editor? Since it also requires changes to defaultDataPropertyForm.ftl I think it should be in a followup PR.

awoods:
> > @gneissone : Is there value in additionally enabling a "textarea" option for larger textual content, per [#76 (review)](https://github.com/vivo-project/Vitro/pull/76#pullrequestreview-137568614) ?
> 
> Yes, I think so. Do you have a suggestion for what value of the editing property should result in a bigger plaintext editor? Since it also requires changes to defaultDataPropertyForm.ftl I think it should be in a followup PR.

A follow-up ticket/PR makes sense. I would defer to the ontologists for the term.

awoods:
Follow-on ticket, per: https://github.com/vivo-project/Vitro/pull/138#issuecomment-603269702
https://jira.lyrasis.org/browse/VIVO-1757

PR #164, opened 2020-05-20
[VIVO-1872] - Add download option to SPARQL Query page

gneissone:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1872)**: VIVO-1872

Adds a new 'force download' check box to the SPARQL query page.

Screen Shot 2020-05-20 at 4 39 53 PM

# What's new?

- Adds new 'force download' checkbox to page
- Changes response headers to force a file download
- Adds an extension to the file based on the format selected 

# How should this be tested?

1. Build with changes. 
2. Go to Site Admin -> SPARQL query.
3. Verify new button is present.
4. Check 'force download' box.
5. Try different formats. Confirm file is downloaded and has the correct filename extension.
6. Uncheck box. 
7. Confirm file does not download (NOTE: this is dependent on your browser support for the requested format. For example, my browser doesn't support RDF/XML so this format is downloaded regardless of whether the 'force download' box is checked)

# Additional Notes:
The motivation for this PR was frustration with very large SPARQL query results causing my browser to hang. 

I don't think this has any implications for SPARQL queries submitted via API, though it would be worth checking. I did test the API and my query was successful. 

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/admin/SparqlQueryController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/admin/SparqlQueryController.java
@@ -40,6 +40,8 @@ import edu.cornell.mannlib.vitro.webapp.rdfservice.RDFServiceException;
 import edu.cornell.mannlib.vitro.webapp.utils.http.AcceptHeaderParsingException;
 import edu.cornell.mannlib.vitro.webapp.utils.http.NotAcceptableException;
 import edu.cornell.mannlib.vitro.webapp.utils.sparql.SparqlQueryUtils;
+import edu.cornell.mannlib.vitro.webapp.controller.api.sparqlquery.RdfResultMediaType;
+import edu.cornell.mannlib.vitro.webapp.controller.api.sparqlquery.ResultSetMediaType;
 
 /**
  * Present the SPARQL Query form, and execute the queries.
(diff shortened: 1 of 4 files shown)


gneissone:
Side note, I did notice that the strings in admin-sparqlQueryForm.ftl are hard coded into the template in English. I checked for alternate versions or properties strings in other languages but it doesn't look like it's been done. Maybe a task to add to the current i18n sprint.

awoods:
Would you be kind enough to create a ticket, @gneissone . If so, I will work it this sprint.

gneissone:
@awoods https://jira.lyrasis.org/browse/VIVO-1873

gneissone:
Is 'force download of results' awkward? Open to suggestions. 
'force downloading of results'
'force results to download'
'force download'
'download results'  <- I think this is probably the best but could be confusing since results will sometimes download anyway without the box checked

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/admin/SparqlQueryController.java
```diff
@@ -110,11 +112,20 @@ private void respondToQuery(HttpServletRequest req, HttpServletResponse resp)
...
-			String format = interpretRequestedFormats(req, queryString);
+			String format = interpretRequestedFormats(req, query);
 			SparqlQueryApiExecutor core = SparqlQueryApiExecutor.instance(
 					rdfService, queryString, format);
 			resp.setContentType(core.getMediaType());
+
+			if (download) {	
+				String extension = getFilenameExtension(req, query, format);
+				resp.setHeader("Content-Transfer-Encoding", "binary");
+				resp.setHeader("Content-disposition", "attachment; filename=sparqlquery." + extension);
```
awoods, line 126: This looks and works well. My only minor suggestion is to potentially change the downloaded filename to "query-results" ... instead of "sparqlquery".

gneissone: Sure, I just went with sparqlquery because that's what the filename was when the browser can't support the file type.

gneissone:
@awoods incorporated your suggestions, plus changed text on query page to `save results to file`

PR #165, opened 2020-05-21
[VIVO-1839] - Use language name for selector instead of flags

gneissone:
**[JIRA Issue](https://jira.duraspace.org/projects/VIVO)**: https://jira.lyrasis.org/browse/VIVO-1839

From Slack 
Screen Shot 2020-05-21 at 1 08 38 PM

Changes the language selector dropdown from displaying flag images to displaying the locale name.

Screen Shot 2020-05-21 at 12 56 37 PM Screen Shot 2020-05-21 at 12 56 52 PM

# What's new?
Displays the locale name in the language selector (in that locale's language, e.g. the text for German will always say `Deutsch` regardless of which language is currently selected. 

# How should this be tested?
Build with changes. Confirm language selector still works as intended and languages are display as text, not images. 

# Additional Notes:
The label is generated from Java's Locale library, so the language must use a recognized locale code to work correctly. This matches the current behavior, which only made an exception for a special 'es-GO' translation that used to be included with VIVO (see https://github.com/vivo-project/Vitro/blob/6d64d27fe4ad26bee7ffd10a852e71240f8ae903/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/selection/LocaleSelectionSetup.java#L143). 

# Interested parties
@awoods


gneissone:
If this is merged, a handful of flag images can be deleted from VIVO, VIVO language and Vitro language repos. 
Screen Shot 2020-05-21 at 12 58 53 PM
Screen Shot 2020-05-21 at 12 59 28 PM
Screen Shot 2020-05-21 at 12 59 09 PM

awoods:
Actually, with the fix of a trivial merge conflict in the file:
`api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/AdditionsAndRetractions.java`
.. I was able target this pull-request at the `sprint-i18n` branch.
Now that the `sprint-i18n` branch has two different English language locales (en_US and en_CA), it surfaces the fact that they both show up as "English" in the dropdown.

Given this, if you don't mind, I would like:
1. to have this PR targeted at the `sprint-i18n` branch, and 
2. to update the PR to include both the language and locale.

Thoughts?

gneissone:
What do you suggest is the default behavior? Always show place name?

'English (United States)'
'English (Canada)'
'Français (Canada)' ?

Or do we want the place only if there are multiple of the same? Will sites want to show multiple versions of the same language?

gneissone:
Based on comment on Slack, updated to include country name. 
Screen Shot 2020-05-22 at 4 51 38 PM

webapp/src/main/webapp/templates/freemarker/page/partials/languageSelector.ftl
```diff
@@ -1,18 +1,12 @@
...
- -->
-
 <#-- This is included by identity.ftl  -->
 <#if selectLocale??>
 <li><ul class="language-dropdown">  <li id="language-menu"><a id="lang-link" href="#" title="user">${i18n().select_a_language}</a><ul class="sub_menu">
     <#list selectLocale.locales as locale>
 
             <li <#if locale.selected>status="selected"</#if>>
-                	<a href="${selectLocale.selectLocaleUrl}?selection=${locale.code}" title="${i18n().select_locale} -- ${locale.label}"><img src="${locale.imageUrl}" title="${i18n().select_locale} -- ${locale.label}" height="15" style="vertical-align:middle" alt="${locale.label}"/></a>
+                	<a href="${selectLocale.selectLocaleUrl}?selection=${locale.code}" title="${i18n().select_locale} -- ${locale.label}">${locale.label}<#if locale.country?has_content> (${locale.country})</#if></a>
```
awoods, line 9: Please also update the comment below that includes mention of `imageUrl` (line 23 in languageSelector.ftl)

gneissone:
> I was unable to see the changes introduced here because this PR branch is not current with the `sprint-i18n` branch. Specifically, the versions in the pom.xml files all indicate "1.11.1-SNAPSHOT" instead of "1.11.2-SNAPSHOT"... which is not picked up by VIVO during build.

An unfortunate side-effect of jumping between pointing the pull request at master and i18n-sprint. It will be nice to get master merged into the sprint branch. Out of curiosity, @awoods did you check out my branch or are you checking out the pull request via `git fetch origin pull/ID/head:BRANCHNAME`? I wonder if the latter method avoids this issue.

awoods:
I checked out the `pull/ID`

gneissone:
Right, I just rebased my branch back in time to a point before there were merge conflicts between master and sprint-i18n, saw there were only relevant files in my diff on the pull request and called it a day. I should have merged the current sprint-i18n branch back in, which I have now done.

awoods:
This looks good... and we can move forward as it stands. I will note, however, that the dropdown items would render better if the width were expanded.
See: 
Screenshot from 2020-05-27 12-52-18

@gneissone , thoughts?

gneissone:
I already pushed out the min-width by 40 pixels so it looked good on my screen, didn't try other resolutions. I wonder if it's necessary to set the width at all, or if it can be as wide as the longest list item.

gneissone:
@awoods Changed the min-width to be an em value, see if that works for you. If not, please suggest a width that looks okay.

webapp/src/main/webapp/css/vitro.css
```diff
@@ -255,7 +255,7 @@ ul.language-dropdown {
 }
 ul.language-dropdown li#language-menu {
     background: url(../images/arrowDownOverAccount.gif) right 9px no-repeat;
-    min-width:150px;
+    min-width:20em;
```
awoods, line 258: This does not result in a fix for me. The width of the dropdown seems to be the same size required by the text: 
- "Select a language"

I still render the same as the previous screenshot. It would be nice to have a fix here... but if it is not worth the time, we can move forward.

awoods: Here is a trade-off that works for the dropdown, but moves the unexpanded menu slightly to the left:
```
 ul.language-dropdown {
     position: relative;
+    min-width: 160px;
 }
 ul.language-dropdown li#language-menu {
     background: url(../images/arrowDownOverAccount.gif) right 9px no-repeat;
-    min-width:20em;
     position: relative;
     margin-right: -10px !important;
 }
```

gneissone: Screen Shot 2020-05-28 at 10 30 33 PM

This results in the menu appearing to jump to the left when opening it, is that what you see?

gneissone: Does this work for you?

```
ul.language-dropdown {
    position: relative;
    min-width: 160px;
}
ul.language-dropdown li#language-menu {
    background: url(../images/arrowDownOverAccount.gif) right 9px no-repeat;
    position: relative;
    margin-right: -10px !important;
    min-width: 170px;
}
```

Screen Shot 2020-05-28 at 10 33 53 PM

awoods: @gneissone : your suggested css update renders nicely. Let's go with it.

nicalico:
Great improvement! However, capitalizing is not consistent. I suggest to change 

`<a href="${selectLocale.selectLocaleUrl}?selection=${locale.code}" title="${i18n().select_locale} -- ${locale.label}">${locale.label}<#if locale.country?has_content> (${locale.country})</#if></a>`

to 

`<a href="${selectLocale.selectLocaleUrl}?selection=${locale.code}" title="${i18n().select_locale} -- ${locale.label}">${locale.label?capitalize}<#if locale.country?has_content> (${locale.country})</#if></a>`

webapp/src/main/webapp/css/vitro.css
nicalico: Gentlemen, it seems like the fix works in Chrome but in Firefox.

image

awoods: @nicalico: do you have any interest in tweaking the css to make this work?
Otherwise, I would suggest we move this change forward, and refine later.

nicalico: I could try to tweak it, but it would have to wait a few days.

gneissone:
@nicalico I incorporated your capitalization suggestion, thanks. I don't have a nice cross-browser solution for the dropdown width, but I'm open to suggestions or they could be submitted as part of a different pull request.

nicalico:
Thanks for the capitalizing. As for the Firefox fix, I have little time to work on it at the moment -- not to mention my css-fu isn't outstanding -- so maybe we should just move the PR forward as suggested by @awoods, and reopen the ticket when a fix occur to us.

nicalico:
@gneissone I have a hard time untangling the commits, here. Did you change something to the css? It doesn't look so, and yet the menu now displays correctly in FF.

gneissone:
@nicalico Perhaps its easier to just look at the [file diff](https://github.com/vivo-project/Vitro/pull/165/files) instead of the commit chain, since adjustments were made to the css multiple times based on feedback. 

I didn't change anything in the css in the last commit. I changed the dropdown to be slightly wider on May 29, though.

nicalico:
@gneissone Yes, I saw that -- but I reported the FF bug on June 6th. Beats me. At any rate, I pulled / recompiled 30 minutes ago, and it looks perfect. Good to merge, in my opinion.

PR #169, opened 2020-06-24
Enable lang selection without need for 'available-langs.properties'

awoods:
## Related PRs:
- https://github.com/vivo-project/VIVO-languages/pull/57
- https://github.com/vivo-project/Vitro-languages/pull/24
- https://github.com/vivo-project/VIVO/pull/174

Resolves: https://jira.lyrasis.org/browse/VIVO-1848

This pull-request, in combination with the three other pull-requests noted below, does the following:
- makes languages available to be enabled without the need for the previous 'available-langs.properties' file

# How should this be tested?
Use the same tests as those documented in the previous pull-request: https://github.com/vivo-project/Vitro/pull/160

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/migration/rel18/FauxPropertiesUpdater.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/migration/rel18/FauxPropertiesUpdater.java
@@ -27,7 +27,7 @@ import edu.cornell.mannlib.vitro.webapp.utils.jena.criticalsection.LockableOntMo
 import edu.cornell.mannlib.vitro.webapp.utils.jena.criticalsection.LockedOntModel;
 
 /**
- * PropertyConfig.n3 has moved to rdf/display/firsttime, so it is only pulled in
+ * PropertyConfig.n3 has moved to rdf/core/display/firsttime, so it is only pulled in
  * if the display model is empty.
  *
  * Let's pull it in one time, anyway.
@@ -39,7 +39,7 @@ import edu.cornell.mannlib.vitro.webapp.utils.jena.criticalsection.LockedOntMode
  * firsttime. If we can't find the file, warn
  */
 public class FauxPropertiesUpdater {
-	private static final String[] PATH_TO_PROPERTY_CONFIG = { "rdf", "display",
+	private static final String[] PATH_TO_PROPERTY_CONFIG = { "rdf", "core", "display",
 			"firsttime", "PropertyConfig.n3" };
 
 	private static final Resource DISPLAY_MODEL = createResource(ModelNames.DISPLAY);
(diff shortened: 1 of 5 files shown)


awoods:
@gneissone : to be clear, for example, are you suggesting the changes to [home/src/main/assembly/home.xml](https://github.com/vivo-project/Vitro/pull/169/files#diff-3b24fb8f9b679a01f25125345f987452) be reverted, and the files in [home/src/main/resources/rdf](https://github.com/vivo-project/Vitro/tree/master/home/src/main/resources/rdf) be moved into a new directory: `home/src/main/resources/rdf/core`?

gneissone:
> @gneissone : to be clear, for example, are you suggesting the changes to [home/src/main/assembly/home.xml](https://github.com/vivo-project/Vitro/pull/169/files#diff-3b24fb8f9b679a01f25125345f987452) be reverted, and the files in [home/src/main/resources/rdf](https://github.com/vivo-project/Vitro/tree/master/home/src/main/resources/rdf) be moved into a new directory: `home/src/main/resources/rdf/core`?

Yes, exactly. And likewise in VIVO. I think it makes more sense than the files mysteriously moving into a new directory during mvn install.

awoods:
The same trickier is also happening in in VIVO-languages:
https://github.com/vivo-project/VIVO-languages/pull/57/files

Are you comfortable with that?

gneissone:
Yeah, I think if it's changed here and in Vitro and it should also be changed in VIVO-languages. 
I guess the overall question is if it will be more obvious for people with existing installer directories (3rd tier) to add the maven black magic to their home.xml file or create a new core directory. It seems more logical to me that the contents of home/src/main/resources are dropped wholesale into home, rather than this specific directory getting special treatment.

awoods:
@gneissone : I am happy to make the change of moving the location of files around inside the `resources` directory... if you/we are generally in favor of the more complex `rdf` directory structure being introduced in this PR.

gneissone:
It's fine with me. I guess there are a few alternatives I can think of...

1.  Continue to put all language-agnostic/file system files into a core directory, load all of these every time. Return to old file structure otherwise and use filtering logic on these.
2. Return to old directory structure, filter out all files with an underscore if there isn't an enabled language suffix following the underscore. 
3. Return to old directory structure, filter out all files with a valid locale (based on Java library) that are not enabled.

I suppose the least disruptive would be options 2 or 3 since custom installer overlays would not need to be reorganized to upgrade. Will be interested to hear other opinions because it's not a big deal to me.

awoods:
Thanks, @gneissone . I appreciate your articulation of options.
I will wait to hear from others: @vivo-project/vivo-committers ... the community.

gneissone (approved):
The new directory structure works for me
```
rdf
├── abox
│   ├── filegraph
│   └── firsttime
├── applicationMetadata
│   └── firsttime
├── auth
│   └── everytime
├── display
│   ├── everytime
│   └── firsttime
├── displayDisplay
│   └── everytime
├── displayTbox
│   └── everytime
├── i18n
│   ├── de_DE
│   │   ├── applicationMetadata
│   │   │   └── firsttime
│   │   ├── display
│   │   │   └── firsttime
│   │   └── tbox
│   │       ├── everytime
│   │       ├── filegraph
│   │       └── firsttime
└── tbox
    ├── filegraph
    └── firsttime
```

I have concerns about receiving missing string errors on initial startup and with annotations being in the 'everytime' directory, but those both appear to have been introduced in previous commits and are unrelated to this. Additionally, we need to be sure that files in the home directory that have been moved at some point in the i18n sprints (e.g. https://github.com/vivo-project/VIVO/blob/master/home/src/main/resources/rdf/display/firsttime/aboutPage.n3) are well-documented in the upgrade documentation.

PR #174, opened 2020-07-13
Add parent reference to installer/pom.xml

awoods:
Resolves: https://jira.lyrasis.org/browse/VIVO-1903

This pull-request updates the `installer/pom.xml` to include a reference to its parent pom.xml.

# How should this be tested?
1. Prior to applying this pull-request, remove all files/directories from your local ~/.m2/repository/org/vivoweb/ directory, followed by a standard VIVO build (mvn clean install -s your-settings.xml)
   - BOOM! build fails
1. Apply this pull-request, and the above process should succeed.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/installer/pom.xml
+++ b/installer/pom.xml
@@ -10,6 +10,13 @@
     <version>1.11.2-SNAPSHOT</version>
     <packaging>pom</packaging>
 
+    <parent>
+        <groupId>org.vivoweb</groupId>
+        <artifactId>vitro-project</artifactId>
+        <version>1.11.2-SNAPSHOT</version>
+        <relativePath>..</relativePath>
+    </parent>
+
     <name>Vitro Installer</name>
 
     <properties>


grahamtriggs:
The installer pom.xml is deliberately written without a reference to parent as it is designed to be used as a standalone project ("binary" install), although it can be and is often used chained into the overall build to allow deployment of the artifacts of a full source build without installation to a local repository.

It has always been this way since Maven was introduced 4+ years ago. Needs more specific details as to why this is now failing - is it specific to a change in / versions of Maven, or is it a change in the installer poms that now prevents them being used independently (which then has knock on consequences of requiring everyone to do full source builds rather than using the artifacts deployed to Maven Central).

awoods:
Hi @grahamtriggs , Vitro/installer/webapp now depends on `vitro-languages` artifacts. 
https://github.com/vivo-project/Vitro/blob/sprint-i18n/installer/webapp/pom.xml#L151-L156

We can either add its own `<distributionManagement>` block to the webapp's pom.xml, such as:
https://github.com/vivo-project/Vitro/blob/sprint-i18n/pom.xml#L457-L470

..or we can follow the standard pattern of the other Vitro / VIVO modules and inherit from the pom.xml of their parent module. The latter approach is taken in this PR.

PR #184, opened 2020-08-31
[VIVO-1915] i18n - Adding a label with language tag through the "new" manageLabelsForindividualAddForm.ftl does not take the setting for language tag

dofeldsc:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1915)**

related PRs:
[1 PR ](https://github.com/vivo-project/VIVO/pull/190)
[2 PR](https://github.com/vivo-project/Vitro-languages/pull/31)

Fixes the bug when adding a label with language tag through manageLabelsForindivdualAddForm.
The selected language-tag was not used, in the end the parent language selection was used.

# How should this be tested?
With a fresh VIVO instanz and the given sample data (https://wiki.lyrasis.org/display/VIVODOC19x/Sample+Data) you can added a label/title to an instanz in an different language (f.e. german). Now you can click the button right of the label to "manage labels". You see the first initial english label and a german label.

If you try to add another label (the "main" language selection is set to english, upper right corner) with the button "Add Label" and you choose for example "French (Canada)" and click "Save". The Label will be listed as english (see images). And now, with the fix, the new label should have the selected language.

# Additional Notes:
I have also tested a few other places and found no problems but maybe someone with more experience in this area of the software should look at it and tell me if this is the right way to fix it.

### DIFF
--- a/webapp/src/main/webapp/js/individual/manageLabelsForIndividual.js
+++ b/webapp/src/main/webapp/js/individual/manageLabelsForIndividual.js
@@ -35,7 +35,6 @@ var manageLabels = {
     // Initial page setup. Called only at page load.
     initPage: function() {
 
-        //var disableSubmit = true;
         if(this.submissionErrorsExist == "false") {
         	//hide the form to add label
         	this.addLabelForm.hide();
@@ -44,7 +43,11 @@ var manageLabels = {
         	if(this.numberAvailableLocales == 0) {
             	manageLabels.showFormButtonWrapper.hide();
             	this.showCancelOnlyButton.show();
-        	} else {
+        	} else if(manageLabels.localeEntryExisting == "true") { 
+        		// if there is already an label for the selected langauge hide the add button 
+        		manageLabels.showFormButtonWrapper.hide();
+        		this.showCancelOnlyButton.show();
+        	} else{
         		//if the add label button is visible, don't need cancel only link
         		this.showCancelOnlyButton.hide();
         	}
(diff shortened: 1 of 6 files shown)


gneissone:
This is working well with adding new labels, but it is not honoring locales when editing labels through the same interface. It appears that clicking 'Edit' brings the user to a form that uses RDFSLabelGenerator which is not aware of the labels current locale, and instead creates a new label with the currently active locale.

awoods:
@gneissone : On today's sprint wrap-up call the team decided that the resolution to this ticket should be to remove the in-form language dropdown for the edit-label page... and as a result, rely on the language context set at the site-level to determine the langTag for added labels.

dofeldsc:
I removed the language select in both files but not removed the earlier fix with selected language.
This is because there is a fr-CA version of the manageLabelsForIndividual.ftl (https://github.com/vivo-project/Vitro-languages/blob/sprint-i18n/fr_CA/webapp/src/main/webapp/templates/freemarker/body/individual/manageLabelsForIndividual_fr_CA.ftl) 

So i left the functionality (mainly java script) in since the fr-CA needs it and we had said that it should probably be reinstalled at a later time.
I had to comment out some part of the functionality (Submit button disable) because it was linked to the language selection but now if no language is selected and the button is pressed anyway, the selected language is used. So understandable behaviour from my point of view.

The second selection is removed via this [PR](https://github.com/vivo-project/VIVO/pull/190)

Does anyone have an different opinion on that?

awoods:
I would suggest we remove:
```
./Vitro-languages/fr_CA/webapp/src/main/webapp/templates/freemarker/body/individual/manageLabelsForIndividual_fr_CA.ftl
```
..as it is functionally identical to the corresponding file in `Vitro`:
```
diff -b ./Vitro-languages/fr_CA/webapp/src/main/webapp/templates/freemarker/body/individual/manageLabelsForIndividual_fr_CA.ftl ./Vitro/webapp/src/main/webapp/templates/freemarker/body/individual/manageLabelsForIndividual.ftl
```

If the "fr_CA.ftl" file is removed, would that allow for removing the "earlier fix with selected language"?

webapp/src/main/webapp/js/individual/manageLabelsForIndividual.js
```diff
@@ -54,37 +54,37 @@ var manageLabels = {
         	this.onShowAddForm();
 
         	//Also make sure the save button is enabled in case there is a value selected for the drop down
-        	if(this.labelLanguage.val() != "") {
-        		disableSubmit = false;
-        	}
+        	// if(this.labelLanguage.val() != "") {
```
awoods, line 57: Instead of commenting out, please remove the code. We can always find it in the history.

awoods (changes requested):

awoods:
> The other thing with editing existing labels: If you edit an for example english label from the list and the language (global) is set to german and you save it, the original label is still there with english language tag and the new edited label gets the german language tag. Should we leave it like this?

This seems like a usability issue... but one that we can address separately from the current issue. I have created a new ticket for this.
https://jira.lyrasis.org/browse/VIVO-1925

@dofeldsc : if you are interested in working this new JIRA, that would be great.

webapp/src/main/webapp/js/individual/manageLabelsForIndividual.js
dofeldsc: Okay, i will remove this file and yes, the earlier fix can then be removed.
I will remove the JS-code too

dofeldsc:
I found a functionality that prevented you from adding two labels in the same language (using the dropdown). I adjusted the functionality so that it still works without the dropdown. The Button "Add Label" is only visible if the selected langauge has no label.

PR #188, opened 2020-09-25
[VIVO-1925] i18n: Editing labels results in new label

dofeldsc:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1925)**

Fixes a bug when you editing language labels.

# How should this be tested?
see [JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1925)
**Bug description from JIRA Issue:
T**his bug can be seen when editing a label where the language of the label is different than the site language context.

For example, if the sample-data (https://github.com/vivo-project/sample-data/blob/master/sample-data.n3) is loaded, add a new label for "Rhetoric" while in a language context other than English (e.g. German).
http://localhost:8080/vivo/editRequestDispatch?subjectUri=http%3A%2F%2Fvivo.mydomain.edu%2Findividual%2Fn2854&editForm=edu.cornell.mannlib.vitro.webapp.edit.n3editing.configuration.generators.VIVOManageLabelsGenerator&predicateUri=http%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23label

Then, switch the site language context to English, and click on the "Edit" button next to the new, German label. Edit the value and save.

Notice the bug: the German label was not actually edited, but rather, a new label was added with an English langTag.

**And now with the fix, the selected label will be changed no matter which language is selected.**

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
@@ -304,10 +304,12 @@ public class MultiValueEditSubmission {
 					String rangeLang = field.getRangeLang();  //UQAM  Default value
 					try {
 						if (_vreq != null ) {
-							if (editConfig.getLiteralsInScope().get("label").get(0).getLanguage() != "")
+							// if the language is set in the given Literal, this language-tag should be used and remain the same
+							// for example when you edit an label with an langauge-tag (no matter which language is selected globally)
+							if (!StringUtils.isBlank(editConfig.getLiteralsInScope().get("label").get(0).getLanguage()))
 							{
 								rangeLang = editConfig.getLiteralsInScope().get("label").get(0).getLanguage();
-							} else {
+							} else { // if the literal has no langauge-tag, use the language which is globally selected
 								rangeLang = _vreq.getLocale().getLanguage();
 								if (!_vreq.getLocale().getCountry().isEmpty()) {
 									rangeLang += "-" + _vreq.getLocale().getCountry();
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
@@ -304,9 +304,14 @@ public class MultiValueEditSubmission {
 					String rangeLang = field.getRangeLang();  //UQAM  Default value
 					try {
 						if (_vreq != null ) {
+							// only if the request comes from the rdfsLabelGenerator the language should be used
+							Boolean get_label_language = false;
+							if (!StringUtils.isBlank(editConfig.formUrl) && editConfig.formUrl.contains("RDFSLabelGenerator")) {
+								get_label_language = true;
+							}
 							// if the language is set in the given Literal, this language-tag should be used and remain the same
 							// for example when you edit an label with an langauge-tag (no matter which language is selected globally)
-							if (!StringUtils.isBlank(editConfig.getLiteralsInScope().get("label").get(0).getLanguage()))
+							if (!StringUtils.isBlank(editConfig.getLiteralsInScope().get("label").get(0).getLanguage()) && get_label_language)
 							{
 								rangeLang = editConfig.getLiteralsInScope().get("label").get(0).getLanguage();
 							} else { // if the literal has no langauge-tag, use the language which is globally selected
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
@@ -21,6 +21,7 @@ import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.ModelFactory;
 import org.apache.jena.rdf.model.ResourceFactory;
 import org.apache.jena.vocabulary.XSD;
+import org.apache.jena.vocabulary.RDF;
 
 import edu.cornell.mannlib.vitro.webapp.controller.VitroRequest;
 import edu.cornell.mannlib.vitro.webapp.edit.EditLiteral;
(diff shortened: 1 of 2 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
```diff
@@ -303,9 +304,14 @@ public void addLiteralToForm(EditConfigurationVTwo editConfig, FieldVTwo field,
 					String rangeLang = field.getRangeLang();  //UQAM  Default value
 					try {
 						if (_vreq != null ) {
-							rangeLang = _vreq.getLocale().getLanguage();
-							if (!_vreq.getLocale().getCountry().isEmpty()) {
-								rangeLang += "-" + _vreq.getLocale().getCountry();
+							if (editConfig.getLiteralsInScope().get("label").get(0).getLanguage() != "")
```
awoods, line 307: Instead of `!=`, should likely use `!...equals()`.
..or `.isEmpty()`.

Is there a chance that the language could be null? If so, maybe `StringUtils.isBlank()`.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/ProcessRdfForm.java
```diff
@@ -308,9 +311,13 @@ protected AdditionsAndRetractions parseN3ToChange(
             String lingCxt=null;
             //UQAM Taking into account the linguistic context in retract
             try {
-                lingCxt = vreq.getLocale().getLanguage();
-                if (!vreq.getLocale().getCountry().isEmpty()) {
-                    lingCxt += "-" + vreq.getLocale().getCountry();
+                if (editConfig != null && editConfig.getLiteralsInScope().get("label").get(0).getLanguage() != "") {
```
awoods, line 314: Instead of `!=`, should likely use `!...equals()`.
..or `.isEmpty()`.

Is there a chance that the language could be null? If so, maybe `StringUtils.isBlank()`.

awoods (changes requested):
This works as expected. For future maintenance, could you add some inline comments to help explain the logic?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
dofeldsc: good point, i changed it. I don't know if it could be null, so better catch it.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/ProcessRdfForm.java
dofeldsc: good point, i changed it. I don't know if it could be null, so better catch it.

awoods (changes requested):
When testing this functionality, it seems that this PR introduces a regression. With the following steps, I never see the option/icon for adding a new label for a given language context. Every pathway I take results in updating the label for all languages.
Steps:
- stop VIVO
- remove all RDF and TDB files from VIVO_HOME
- ensure multiple languages are configured in runtime.properties
- start VIVO
- load [sample data](https://raw.githubusercontent.com/vivo-project/sample-data/master/sample-data.n3) via site-admin
- Set site language context to en_US
- Navigate to [people](http://localhost:8080/vivo/people)
- Navigate to "Bogart, Andrew"
- Click "Rhetoric" under "Research Areas"
- Click the edit/pencil icon next to the "Rhetoric" label
- Switch the site language context... no matter which language is selected, no "Add label" button appears and any edit of the label applies to all languages.

dofeldsc:
oh yeah, i found the problem. When I developed and tested it, there were already at least two labels with different languages, so I did not see the case. 
When you got at least two labels, you get this icon
manage_labels. This brings you to another page and everything works as intended. 
I have to take a close look at it again, maybe we can talk about it later when you have time.

dofeldsc:
I found a solution but it is more of a workaround then a fix.
The only other option I can think of is to "forbid" direct editing of the label on the profile page and always link to the "Manage Labels" page 
Bildschirmfoto vom 2020-09-30 16-16-44

But this would not be a good solution either.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
```diff
@@ -304,9 +304,14 @@ public void addLiteralToForm(EditConfigurationVTwo editConfig, FieldVTwo field,
 					String rangeLang = field.getRangeLang();  //UQAM  Default value
 					try {
 						if (_vreq != null ) {
+							// only if the request comes from the rdfsLabelGenerator the language should be used
+							Boolean get_label_language = false;
+							if (!StringUtils.isBlank(editConfig.formUrl) && editConfig.formUrl.contains("RDFSLabelGenerator")) {
+								get_label_language = true;
```
awoods, line 310: Please change variable name to camelCase: `getLabelLanguage`.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/ProcessRdfForm.java
```diff
@@ -312,9 +312,14 @@ protected AdditionsAndRetractions parseN3ToChange(
             String lingCxt=null;
             //UQAM Taking into account the linguistic context in retract
             try {
+                // only if the request comes from the rdfsLabelGenerator the language should be used
+                Boolean get_label_language = false;
```
awoods, line 316: Please change variable name to camelCase: `getLabelLanguage`.

awoods (changes requested):
Works great. A minor comment on syntax, then this should be ready for merge. Thanks.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
dofeldsc: done with commit 47196383fe1fefb5da843e985c4d2f5917d14696

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/ProcessRdfForm.java
dofeldsc: done with commit 47196383fe1fefb5da843e985c4d2f5917d14696

awoods (approved):

PR #192, opened 2020-10-17
Create vivo/home on application start-up

awoods:
- Add logic for untarring vivo-home

See details on related PR: https://github.com/vivo-project/VIVO/pull/197

Related to: https://jira.lyrasis.org/browse/VIVO-1443

# Interested parties
@VIVO-project/vivo-committers


wwelling (dismissed):

gneissone:
@wwelling Testing out the new digest logic. Seems like
 - VIVO explodes tar every startup
 - VIVO deploys all files in tar, except those that have been modified directly in vivo-home.
    -> Thus updates made directly to vivo-home supercede updates made in the custom-webapp

I _think_ this is the expected behavior, though it is a reversal (fix?) of the prevailing pattern in VIVO where if you mvn install'd it would overwrite the file in vivo-home, regardless if it had been updated.

Can we avoid deploying all the unchanged files, too? I.e., only execute the deploy logic if vivo-home has changed between two subsequent starts? Is that tracked in digest.md or would it require a second checksum?

awoods:
@gneissone, can you be more specific with your expected behavior? I am not exactly sure I understand what you would like with the suggestion:
> only execute the deploy logic if vivo-home has changed between two subsequent starts

gneissone:
It seems superfluous for VIVO to overwrite files in vivo home if they are exactly the same between the tar'd up home and the vivo home as it was in the previous startup. I can tell they are being overwritten because the 'Date Modified' date changes on every startup.

wwelling:
@gneissone I suppose the files do not even need to be written if the checksum matches.

wwelling:
@gneissone the digest is the original checksums of the untarred home directory after the first startup. In subsequent startups, the files are untarred and first checked to see if they already exist in the home directory. If they do, it checks to see if the digest checksum matches the checksum of the current file in the home directory. This is essentially determining if the file in the home directory has changed since the last time the digest was generated.

If we do not want to overwrite with a newly untarred file that already exists in the home directory and has not changed, we would have to do another checksum comparison between the newly untarred file and the existing file. We do not currently do this check and just assume to overwrite. We do already perform the checksum and could easily add this logic. It would be the exact same results just without new timestamps.

wwelling:
I will update PR shortly to not overwrite file with exact same file. Otherwise, @gneissone was there anything else not working as expected or not doing something is should?

wwelling:
https://github.com/awoods/Vitro/pull/3

awoods:
We will want to squash the commits in this PR when merging into `main`.

gneissone:
Tested latest iteration and it works as expected. A couple small requests to the log behavior: https://github.com/awoods/Vitro/pull/4

gneissone (approved):

wwelling (approved):

roflinn (approved):

PR #193, opened 2020-10-20
VIVO-1752: Initialize Solr Core on startup

wwelling:
[VIVO-1752](https://jira.lyrasis.org/browse/VIVO-1752)

@VIVO-project/vivo-committers


api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/solr/SolrSearchEngine.java
```diff
@@ -199,13 +233,97 @@ public int documentCount() throws SearchEngineException {
...
+			.map(NamedList.class::cast)
+			.map(status -> status.get("name"))
+			.isPresent();
+	}
+
+	private void initializeCore(String contextPath) throws SolrServerException, IOException {
+		Path solrConfPath = Paths.get(contextPath + File.separator + "solr");
+		File solrConf = solrConfPath.toFile();
+
+		solrConf.setReadable(true, false);
```
awoods, line 292: These are both returning `false`... not surprisingly.

awoods: We may need to document a prerequisite of setting permissions on the Solr directories before start-up.

awoods: Both users need to be able to write to the core directories: 
- solr user
- tomcat user

wwelling: I tried changing the owner and that resulted in AccessDeniedException. I am open to ideas with permission issues.

wwelling: At the point of requiring changing directory permissions, it is not much more to copy the conf directory and create the core.

awoods:
@wwelling : we can probably close this PR.

PR #194, opened 2020-10-30
Reduce number of times QuerySolutions are looped in LanguageFiltering…

awoods:
…RDFService

- No functional change in this update

Resolves: https://jira.lyrasis.org/browse/VIVO-1931

This update changes the amount of time to filter language QueryResults from O(n^2) to O(n)

# How should this be tested?
See JIRA ticket for test scenario and script for producing large, i18n test set
The application should work exactly as before... but faster with large numbers of triples to filter.

# Interested parties
@VIVO-project/vivo-committers and @matthiasluehr


brianjlowe (approved):
Works great.  Very noticeable improvement in page render time, even with more modest numbers of related individuals.

PR #197, opened 2020-11-27
Prevent 'ProcessRdfForm.parseN3ToRDF' from using null linguisticContext

awoods:
Resolves: https://jira.lyrasis.org/browse/VIVO-1944

See JIRA ticket for description of issue and testing procedures.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/ProcessRdfForm.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/ProcessRdfForm.java
@@ -346,7 +346,12 @@ public class ProcessRdfForm {
      */
     protected static List<Model> parseN3ToRDF(
             List<String> n3Strings, N3ParseType parseType, String linguisticContext ) throws Exception {
-       List<String> errorMessages = new ArrayList<String>();
+        // Use non-linguistic version of this method if no linguisticContext is provided
+        if (linguisticContext == null) {
+            return parseN3ToRDF(n3Strings, parseType);
+        }
+
+        List<String> errorMessages = new ArrayList<String>();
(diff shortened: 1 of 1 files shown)


gneissone:
Confirmed that weird things happen when editing a page title. 
Screen Shot 2021-01-05 at 2 47 32 PM

I'm also seeing the problem when language filtering is turned on.
Screen Shot 2021-01-05 at 2 44 21 PM


@awoods Does this affect editing elsewhere in the application, or just with page editing? The fix seems to work, just confirming we aren't preempting a different expected behavior. 

Side note, seems like a seemingly random page label shows up after clicking through to the edit page, ignoring the language context. Editing the label results in more weirdness. Clicked through to edit 'Personen' while in the German context and was presented with the french label. Edited it and ended up with a crazy mixture of them all. Doesn't seem related to this PR, but just wanted to note.

Screen Shot 2021-01-05 at 3 58 10 PM

awoods:
@gneissone : yes, I have seen the same weirdness with regards to seemingly random values showing up on edit pages.

This [method](https://github.com/vivo-project/Vitro/blob/5e9fe6c5953f2fb1b6fec9c406868e23f7bacd85/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/ProcessRdfForm.java#L347) with the [`linguisticContext`](https://github.com/vivo-project/Vitro/blob/5e9fe6c5953f2fb1b6fec9c406868e23f7bacd85/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/ProcessRdfForm.java#L343) parameter was added in the i18n effort. If the `linguisticContext` is "null", the logic defaults to the legacy method.

gneissone (approved):

PR #203, opened 2021-01-28
Changed hard coded time units to property calls in dateTimeWithPrecision.ftl

nicalico:
Fix to : https://jira.lyrasis.org/browse/VIVO-1877
Related PR : 
- https://github.com/vivo-project/VIVO-languages/pull/89
- https://github.com/vivo-project/Vitro-languages/pull/39

### What does it do
Makes the file universal by removing hard coded time units. Required properties are add in https://github.com/vivo-project/VIVO-languages/pull/89.
The dropdown placeholders are not lowercased as in the original, as it would be incorrect in de_DE.

### How to test it
Browse to **Events > [some event] > Overview >  date/time interval > add** and check labels in the form.
Fix has been tested with fr_CA, en_CA, en_US, pt_BR, de_DE and es contexts.

PR #204, opened 2021-01-29
[VIVO-1953] - Use i18n values for datetime form

gneissone:
**[JIRA Issue](https://jira.duraspace.org/projects/VIVO)**: (please link to issue)

Removes hard-coded english values and uses i18n strings.

# What's new?
Should appear the same as before in English, but now support i18n if the language is switched.

# How should this be tested?
Confirm issue by adding a date time value (note, not an interval, that uses a separate form).
Screen Shot 2021-01-28 at 1 23 40 PM

Deploy changes, switch contexts and confirm i18n strings are being used.

Screen Shot 2021-01-28 at 5 25 01 PM

# Additional Notes:
Requires the strings being proposed here: https://github.com/vivo-project/VIVO-languages/pull/89/files
This PR should not be merged prior.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/webapp/src/main/webapp/templates/freemarker/edit/forms/dateTimeEntryForm.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/edit/forms/dateTimeEntryForm.ftl
@@ -3,13 +3,13 @@
 <fieldset class="dateTime">
   <input type="hidden" id="literal" name="literal" value="${literalValues}" role="input"/>
   <#if datatype?contains("#date") || datatype?contains("Year") >
-    <label for="dateTimeField-year">Year</label>
+    <label for="dateTimeField-year">${i18n()[('label.dateTimeWithPrecision.year_capitalized')]}</label>
     <input class="text-field" name="dateTimeField-year" id="dateTimeField-year" type="text" value="" size="4" maxlength="4" />
   </#if>
   <#if datatype?contains("#date") || datatype?contains("Month") >
-    <label for="dateTimeField-month">Month</label>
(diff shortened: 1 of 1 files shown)


webapp/src/main/webapp/templates/freemarker/edit/forms/dateTimeEntryForm.ftl
```diff
@@ -3,13 +3,13 @@
...
-    <label for="dateTimeField-year">Year</label>
+    <label for="dateTimeField-year">${i18n()[('label.dateTimeWithPrecision.year_capitalized')]}</label>
     <input class="text-field" name="dateTimeField-year" id="dateTimeField-year" type="text" value="" size="4" maxlength="4" />
   </#if>
   <#if datatype?contains("#date") || datatype?contains("Month") >
-    <label for="dateTimeField-month">Month</label>
+    <label for="dateTimeField-month">${i18n()[('label.dateTimeWithPrecision.month_capitalized')]}</label>
     <select name="dateTimeField-month" id="dateTimeField-month" >
-        <option value="">month</option>
+        <option value="">${i18n()[('label.dateTimeWithPrecision.month_capitalized')]}</option>
```
awoods, line 12: This was lower-cased before. Did you intentionally choose to make it capitalized?
Same is true to others below.

gneissone: Just copying the logic in https://github.com/vivo-project/Vitro/pull/203/files. The strings are already capitalized in the i18n file so not really sure if it's necessary for freemarker to capitalize them again, but it seemed nice to be consistent with the other date time form.

awoods (approved):

PR #205, opened 2021-01-29
Issue/vivo 1947

brianjlowe:
https://jira.lyrasis.org/browse/VIVO-1947

This pull request adds safeguards to prevent both the ontology editor and the N3 editing from erroneously deleting property values in languages other than the one associated with the editing VitroRequest.

# What's new?
JenaBaseDao.updateRDFSLabel() has been updated, which in turn uses a new generalized method  JenaBaseDao.updatePlainLiteralValue().  The DAOs now use this method to update properties such as public description, example, short definition, etc.  A new ModelChangePreprocessor called LimitRemovalsToLanguage has been added to eliminate unwanted triples from the N3 editor's retractions model.  This model change preprocessor is added to each update in ProcessRdfFormController.  There are a number of other updates to various classes to use request models instead of context models and to make request-filtering DAO classes visible to the OperationController, which will try to invoke their methods via reflection.

# How should this be tested?
1. Until all labels are moved into firsttime, they are not good candidates for test.  Better to edit a class or property in the ontology editor, or open something like Site Admin and change the copyright text or change the page title on a menu page.  Save your change and then switch to a different language.  Perform the edit again, changing the value in that new language.  Switch between the two languages and observe that not only have both of your changes have been preserved, but any existing values in the remaining languages remain as well.
2. Navigate to a faculty member's profile.  Add a value for the "head of" property, entering a string in the role field.  Save the form.  Then switch languages, re-open the form and enter a new value for the role string.  Switch between languages and observe that the role strings displayed on the profile's list view track the language you have selected.


api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/configuration/preprocessors/LimitRemovalsToLanguage.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public void preprocess(Model retractionsModel, Model additionsModel,
+            HttpServletRequest request) {                 
+        log.debug("limiting changes to " + language);
+        List<Statement> eliminatedRetractions = new ArrayList<Statement>();
+        StmtIterator sit = retractionsModel.listStatements();
+        while(sit.hasNext()) {
+            Statement stmt = sit.next();
+            if(!stmt.getObject().isLiteral()) {
+                continue;
```
awoods, line 63: I am always a little wary of `continue` and `break` in loops, as they make following the logic slightly trickier.
In this particular case, the `continue` can easily be removed with:
```
if (stmt.getObject().isLiteral()) {
    ...
```

If, however, you have a reason that I am not seeing for keeping the `continue`, I defer to your discretion.

awoods (changes requested):
The functionality works in the manual tests I performed.

I have one minor code review comment... otherwise, looks good.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/configuration/preprocessors/LimitRemovalsToLanguage.java
brianjlowe: Yep, it was pretty pointless in this case.  Pushed the change.

awoods (approved):

PR #207, opened 2021-02-03
Merge main back into sprint-i18n in prep for big merge

gneissone:
This will bring the sprint branch back up to speed with the main branch. We are targeting Monday to merge sprint-i18n back into main.

@awoods @wwelling @vivo-project/vivo-committers

PR #208, opened 2021-02-03
Sprint i18n main

awoods:
Refresh of the `sprint-i18n` branch with updates from `main`.

Supersedes: https://github.com/vivo-project/Vitro/pull/207


gneissone (approved):
Looks good, though I wouldn't mind if the conflict resolution included removing some of the additional comments in the Java code that I feel don't provide any meaningful context (or adding to the comments so they do).

PR #212, opened 2021-02-17
Use same fallback locale used by RDFFilesLoader when setting locale o…

brianjlowe:
…n request. Resolves https://jira.lyrasis.org/browse/VIVO-1959

**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1959)**: 

Uses the same fallback locale (currently "en_US") used by the RDFFilesLoader to set the request's locale if there is no list of selectable locales from which to obtain a default.

# What's new?
"en_US" fallback is refactored to SelectedLocale.java.  RDFFilesLoader uses this fallback, as does SelectedLocale.getOverridingLocale().

# How should this be tested?
* Before installing pull request:
* Install VIVO with all language-related properties in runtime.properties commented out (standard monolingual VIVO with no explicit forced locale).
* Set your browser's language preference list (wherever it provides the list of languages/countries that can be moved up and down) so that the topmost locale is something unrecognized by VIVO (e.g. "ro-RO").
* Observe that each i18n property usage on the page displays an error message ("Text bundle 'all' not found").
* Install pull request.
* English-US i18n text should appear on the page instead of the error messages.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/selection/SelectedLocale.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/selection/SelectedLocale.java
@@ -91,7 +91,9 @@ public abstract class SelectedLocale {
 			}
 		}
 
-		return null;
+		Locale fallbackLocale = getFallbackLocale();
+        log.debug("Using fallback locale as default: " + fallbackLocale);
+        return fallbackLocale;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


gneissone:
First look using Docker looks good. I see `2021-02-17 16:41:33,840 WARN  [I18n] Didn't find text bundle 'all'` in the log, which is presumably the Nepali bundle my browser is asking for, then it loads English instead.

wwelling (approved):
Looks good and works for me.

gneissone (approved):

PR #216, opened 2021-03-02
Add subject to reindex in AdditionalURIsForObjectProperties

litvinovg:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1966)**

Adds subject to reindex list after adding or removing object property.

# How should this be tested?
1. Enable debugging for AdditionalURIsForObjectProperties
2. Create two persons
3. Add object property Person1 sameAs Person2
4. In logs AdditionalURIsForObjectProperties should find both persons to reindex.

# Interested parties
@VIVO-project/vivo-committers


brianjlowe (approved):
Looks good and works as advertised.

brianjlowe:
For other testers, an alternative to checking the log messages might be to query for the two individuals directly in the Solr admin interface at :8983 before and after adding the sameAs triple.  Without this pull request, the first individual's timestamp remained unchanged while the second's was updated.  After applying the pull request, both individuals' timestamps are updated at the same time.

brianjlowe:
Merging this as 1-review bug.

PR #221, opened 2021-03-30
[VIVO-1736] - Bugfix for redirecting user to home page after login

gneissone:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1736)**: VIVO-1736

Fixes an issue with external login redirects where they lead to an empty page in some situations.

# What's new?
Removes a method for determining the home page that didn't work. Replaced with a method defined elsewhere in the VIVO code. Also added a single line to the debug log that reports back the URL path that is being used for the redirect.

# How should this be tested?
1. Setup an external authentication scheme, such as Shibboleth or CAS. That's the hard part.
2. Login to VIVO using this external mechanism as a user that is not attached to a profile and that does not have elevated privileges (i.e. is a self-editor)
3. If the user has not logged in before, complete the information VIVO asks for, log out and log back in
4. Confirm upon login that a blank page is shown and the browser url shows /loginExternalAuthReturn.

To confirm fix, repeat above but user should be redirected to homepage as intended.

# Additional Notes:
Pure speculation, but this bug may be limited to VIVO installations where VIVO is served at Tomcat's root, and may be a result of how that was set up.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/LoginRedirector.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/LoginRedirector.java
@@ -23,6 +23,7 @@ import edu.cornell.mannlib.vitro.webapp.beans.UserAccount;
 import edu.cornell.mannlib.vitro.webapp.controller.Controllers;
 import edu.cornell.mannlib.vitro.webapp.i18n.I18n;
 import edu.cornell.mannlib.vitro.webapp.i18n.I18nBundle;
+import edu.cornell.mannlib.vitro.webapp.controller.freemarker.UrlBuilder;
 
 /**
  * A user has just completed the login process. What page do we direct them to?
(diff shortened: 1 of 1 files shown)


gneissone:
There is an interesting bit of code within the deleted method that would allow a redirect to be determined by a postLoginRequest session attribute. However, I cannot find any reference to setting that in VIVO or Vitro, so I suspect it is just a bit of legacy code without any current functional use. I found one old reference to a jsp having the potential to set it: https://culibrary.atlassian.net/browse/NIHVIVO-1237

brianjlowe (approved):
Despite not having an easy way to test external authentication end to end, I'm approving this based on the tested behavior of the two methods for finding the home page URL.  As suspected, they differ only when VIVO is served as the root application.  In that case, the removed method returns "" while UrlBuilder.getHomeUrl() returns the expected "/".

brianjlowe:
Declaring this a single-review bugfix and merging.

PR #224, opened 2021-04-06
Move event listening to RDFServiceGraph level to avoid adding a new r…

brianjlowe:
…edundant listener each time a model is made for an existing RDFServiceGraph. Resolve https://jira.lyrasis.org/browse/VIVO-1976

**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1976)**: (please link to issue)

this pull request do?
A brief description of what the intended result of the PR will be and/or what problem it solves.

- Removes registration of new listener from RDFServiceGraph.createRDFServiceModel()
- Adds event listening to anonymous subclass of SimpleEventManager in RDFServiceGraph.getEventManager().
- Adds unit test that creates multiple models from a single RDFServiceGraph and ensures that the RDFServiceGraph hears an event only once.

# How should this be tested?
Confirming that VIVO builds with a successful RDFServiceGraphTest should be sufficient.  Before applying the changes in RDFServiceGraph.java, RDFServiceGraphTest.java will fail because the RDFServiceGraph hears a single event 100 times.  After applying the rest of the PR's changes, it will hear it exactly once.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/jena/RDFServiceGraph.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/jena/RDFServiceGraph.java
@@ -12,10 +12,10 @@ import java.util.function.Supplier;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-
 import org.apache.jena.graph.Capabilities;
 import org.apache.jena.graph.Graph;
 import org.apache.jena.graph.GraphEventManager;
+import org.apache.jena.graph.GraphListener;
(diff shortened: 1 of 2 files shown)


gneissone:
@brianjlowe Thank you for this PR!!!

Question on testing... can you speak a little as to how the notifyEvent method is used in practice? It seems like perhaps its only used by some of the RDF tools, such as the tool for renaming resources? 

And the listener itself? Any thoughts on what functionality I should be ensuring isn't affected by this change?

litvinovg (approved):
Works good. No new org.apache.jena.rdf.model.impl.ModelListenerAdapter after solr rebuild with this fix.

gneissone (approved):

PR #227, opened 2021-04-07
[VIVO-1736] - Bugfix for redirecting user to home page after login

gneissone:
**[JIRA Issue](https://jira.lyrasis.org/browse/VIVO-1736)**: VIVO-1736

Fixes an issue with external login redirects where they lead to an empty page in some situations.

# What's new?
Alters a method for determining the home page that didn't work, such that a trailing slash is added if the context resolves to a blank string. Removes unused functionality for determining the redirect based on a servlet attribute. Also added a single line to the debug log that reports back the URL path that is being used for the redirect.

# How should this be tested?
1. Setup an external authentication scheme, such as Shibboleth or CAS. That's the hard part.
2. Login to VIVO using this external mechanism as a user that is not attached to a profile and that does not have elevated privileges (i.e. is a self-editor)
3. If the user has not logged in before, complete the information VIVO asks for, log out and log back in
4. Confirm upon login that a blank page is shown and the browser url shows /loginExternalAuthReturn.

To confirm fix, repeat above but user should be redirected to homepage as intended.

# Additional Notes:
Pure speculation, but this bug may be limited to VIVO installations where VIVO is served at Tomcat's root, and may be a result of how that was set up. 
Previously merged, then un-merged (see https://github.com/vivo-project/Vitro/pull/221 and https://github.com/vivo-project/Vitro/pull/226). 

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/LoginRedirector.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/authenticate/LoginRedirector.java
@@ -108,7 +108,9 @@ public class LoginRedirector {
 			throws IOException {
 		try {
 			DisplayMessage.setMessage(request, assembleWelcomeMessage());
-			response.sendRedirect(getRedirectionUriForLoggedInUser());
+			String redirectUrl = getRedirectionUriForLoggedInUser();
+			log.debug("Sending redirect to path: " + redirectUrl);
+			response.sendRedirect(redirectUrl);
 		} catch (IOException e) {
 			log.debug("Problem with re-direction", e);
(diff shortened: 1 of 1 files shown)


brianjlowe (approved):

brianjlowe (changes requested):
I just noticed that the comment for getApplicationHomePage() should probably be even further simplified now that the postLoginRequest logic is removed.  That was the only place the absolute URL came into play (if that attribute was set with a value containing a colon.)

gneissone:
@brianjlowe Just removed the comment since it no longer seemed useful.

brianjlowe (approved):

PR #232, opened 2021-05-28
[VIVO-1983] Limit label management values to current locale

brianjlowe:
[https://jira.lyrasis.org/browse/VIVO-1983](https://jira.lyrasis.org/browse/VIVO-1983)

Must be installed with companion VIVO PR https://github.com/vivo-project/VIVO/pull/240

Shows manage labels icon only if there are two or more labels available with the selected locale filtering.  Shows only the language-appropriate labels for editing, to be in line with the editing of other properties.

# How should this be tested?
1. Create two test individuals: one person, and one non-person.  To each, add a label in each of several labels. (Change label dropdown, click pencil next to label).
2. For each individual, upload an RDF triple adding a second label in one of the languages you already used.  E.g. the test individuals each have 1 English label, 1 French label, and 2 German labels.

In the above example, you should see the manage labels Rolodex only on German; clicking the Rolodex will show you only the two German labels for editing. 

@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/ViewLabelsServlet.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/ViewLabelsServlet.java
@@ -188,8 +188,8 @@ public class ViewLabelsServlet extends FreemarkerHttpServlet{
 
         ArrayList<Literal>  labels = new ArrayList<Literal>();
         try {
-        	//We want to get the labels for all the languages, not just the display language
-            ResultSet results = QueryUtils.getLanguageNeutralQueryResults(queryStr, vreq);
+        	// Show only labels with current language filtering
+            ResultSet results = QueryUtils.getQueryResults(queryStr, vreq);
             while (results.hasNext()) {
                 QuerySolution soln = results.nextSolution();
                 Literal nodeLiteral = soln.get("label").asLiteral();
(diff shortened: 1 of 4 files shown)


gneissone:
Looks good, just one question. This PR removes the 'view all labels' link for non-logged-in users. Why was that done? 
It seems like an edge case that one would actually purposefully add multiple labels in the same language, then want end users to view them. So I suppose it's not a commonly used feature, but a feature nonetheless. 

Screen Shot 2021-06-02 at 12 15 26 PM
View labels page

brianjlowe:
I guess that could be a intended feature.  I was thinking it more likely that multiple labels would be an unintended situation arising from ingest from multiple sources, and the appearance of the extra icon would be an unexpected wart on the public profile page.

brianjlowe:
Phew.  OK, the changes to lib-properties.ftl are reverted.  I'm actually not sure why I had thought I had to add those in the first place, since the original logic only shows the label list if (a) label is editable and there is more than one label or (b) there are labels for more than one language.  So the show labels link should appear in a VIVO that has language filtering turned off and labels in multiple languages exist for the individual.  The foaf-person templates, however, weren't passing the language count variable to the label macro the way individual-vitro.ftl does, so the companion VIVO pull request adds those in so that everything should hopefully now be consistent.

gneissone:
Looks good for logged in users. I'm seeing some unexpected behavior for logged out users. In the situation where there is one label with no language tag, then two with the de-DE tag, I see the rolodex icon and link in both english and german contexts, when I would actually expect to see the pencil icon in english. Does that match what you are experiencing?

brianjlowe:
Perhaps I've misunderstood, but I don't seem to be seeing that.  I created both a Person and an Analysis Service.  I gave each two german labels and one label with no language.  When logged out, I don't see the Rolodex on either one on either English or German.  Logged in gives me the rolodex for German and allows me to see/edit both German labels.

litvinovg (approved):
Works good.

gneissone (commented):
Logged in as admin, everything works as expected. I'm now seeing no rolodex when logged out in the case where filtering and multi-language is enabled, and there are multiple labels in a single language (ie pencil icon all the time). We can revisit if we ever need to support this edge case.

PR #236, opened 2021-06-01
[VIVO-1985] Compare locales of equal rank alphabetically to provide consistent fallback

brianjlowe:
**[https://jira.lyrasis.org/browse/VIVO-1985](https://jira.lyrasis.org/browse/VIVO-1985)**:

Updates the language sorting in LangSort so that languages of the same rank are consistently sorted with respect to one another.  This makes for a consistent selection of fallback language if no values appropriate to the current locale setting are available.

To reproduce, enter a research overview in two different languages.  Switch to a third language and edit reseach overview.  Should (likely) get a 'no match to existing literal' exception.

After applying fix, should be able to edit under the third language without error.

# Interested parties
@VIVO-project/vivo-committers


litvinovg (approved):
Found new problems. More details in https://jira.lyrasis.org/browse/VIVO-1985

PR #247, opened 2021-08-27
Improve MAVEN build Performance

ChenZhangg:
[Parallel builds in Maven 3](https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3) Maven 3.x has the capability to perform parallel builds.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.


wwelling (commented):
I do not see the value. If and when these two scripts become a bottleneck in development or building for deployment we can revisit this. I recommend we close this issue.

PR #250, opened 2021-11-01
Creating Locale with specified Language script

VeljkoMaksimovic:
Adds a LocaleUtility class which expands LocaleUtil from lang3 package. LocaleUtil does not support creation of Locales with script specified (Latin, Cyrillic, etc.). LocaleUtility checks if the string that represents locale contains script. If so, it creates Locale using Locale.Builder class, otherwise calls the LocaleUtil.toLocale() as it was done before.

# What's new?
Now we can specify locales like sr_Latn_RS or sr_Cyrl_RS, which was not the case before.

# How should this be tested?
 Only way to see the difference is to have Locales of same languages but in different scripts.

* Does this change require documentation to be updated? 
    - Yes. Localization documentation should contain example of Locale string representation which specifies the script.
 * Does this change add any new dependencies? 
    - Not external

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,3 +20,4 @@ research and scholarship portal, [VIVO](https://duraspace.org/vivo/).
 
 For more information, contact the [VIVO community](https://duraspace.org/vivo/resources/contact/).
 
+
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/selection/LocaleSelectionController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/selection/LocaleSelectionController.java
@@ -12,6 +12,7 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import edu.cornell.mannlib.vitro.webapp.utils.LocaleUtility;
 import org.apache.commons.lang3.LocaleUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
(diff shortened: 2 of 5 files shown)


litvinovg:
LocaleUtils from lang3 package support creation of locales with variants. 
The first should be language in lower case + _ + country in upper case + _ + variant.
Try LocaleUtils.toLocale("sr_RS_Cyr") 
and LocaleUtils.toLocale("sr_RS_Latn")

VeljkoMaksimovic:
Thanks for the reply.
I tried doing it that way before, but it has few problems. First, Locale does not treat script and variant the same, so although I created two languages, one in Latin and another in Cyrillic, like you suggested, this is the result I get in language selection drop-down menu:
SelectLanguageLang3
Both Languages are shown in Cyrillic since that is the default script for sr_RS locale. Also, since Latin and Cyrillic are not valid variants, I also get these warnings:
WarningLang3
On the other hand, if I use Locale.Builder() to set the script, this is the result:
SelectLanguageBuilder
and the warnings are also gone.


Looking at the source code of LocaleUtils, there seams to be no way to specify Script, although IETF BCP 47 standard  https://tools.ietf.org/search/bcp47#section-2.1 , which Java documentation for Locale cites, specifies a way to denote a script within the language tag.

litvinovg (dismissed):
Works as described.

chenejac (dismissed):
It works as described. The name of the LocaleUtility might introduce confusion, because the Utils suffix is used for some Vitro utility classes. However, naming this class LocaleUtils might introduce importing classes issues, i.e. mixing the newly added class with org.apache.commons.lang3.LocaleUtils. Therefore, there is some sense in keeping this name. I approve this PR.

litvinovg (approved):

PR #252, opened 2021-11-17
[VIVO-1841] - Add POST body support to update api

gneissone:
**[JIRA Issue](https://github.com/vivo-project/VIVO/issues/3429)**: https://github.com/vivo-project/VIVO/issues/3429

* https://groups.google.com/forum/#!topic/vivo-tech/RlPFVvoOenQ

Adds additional features to the SPARQL Update API

# What's new?
Adds support for submitting a SPARQL update request as part of a POST body when the request content type is set to `application/sparql-update` as described in the SPARQL 1.1 protocol here: https://www.w3.org/TR/sparql11-protocol/#update-operation. 

Sending the update via the 'update' parameter is still supported.

Additionally, this adds support for supplying the using-graph-uri and using-named-graph-uri parameters laid out in the above protocol.

# How should this be tested?

General info on using the API is here: https://wiki.lyrasis.org/display/VIVODOC112x/SPARQL+Update+API
Description on how 'USING,' and by extension using-graph-uri works: https://stackoverflow.com/a/46370944/6326524

I created a Postman collection to demonstrate the changes to the API. 
[VIVO PR Testing.postman_collection.json.zip](https://github.com/vivo-project/Vitro/files/7571518/VIVO.PR.Testing.postman_collection.json.zip)

You can test using the Postman collection. You will need to update the email and password parameters for the requests. You can also update the request URL if your VIVO isn't running at http://localhost:8080/vivo. The requests in the collection are:

**VIVO Update Old Style** - demonstrates the previous method of adding data via API. It will load the data into a graph named http://localhost/data/test. You can ensure the data was added by going to http://localhost:8080/vivo/individual?uri=https://scholars.mydomain.edu/individual/journal-old. If you are logged in as admin, you can 'edit this individual' to and click 'Raw Statements with This Resource as Subject' to confirm the graph. 

**VIVO Update Post Body** - demonstrates the new method of adding data via the POST body with the 'Content-Type' header set to 'application/sparql-update.' Confirm at http://localhost:8080/vivo/individual?uri=https://scholars.mydomain.edu/individual/journal-new. 

**VIVO Update Post Body USING** - demonstrates how you could add new triples with the USING key and a SPARQL query. Confirm with either of the existing objects that they have been added to a new graph named http://example.com/new-test-graph and with a new triple. 

**VIVO Update Post Body using-graph-uri** - Does the same as above but via setting the 'using-graph-uri' parameter. This is new functionality.

The resulting triples will look like this:
Screen Shot 2021-11-19 at 9 32 41 AM

# Additional Notes:
This reads the POST body as an InputStream and pass this into the updateFactory as such (see https://github.com/vivo-project/Vitro/pull/252/files#diff-ed2a7ed3e38edc1a8257fdb2ad1f58a0e193ece56b1cfb454eadf61b7ea6138dL93 for the change that makes this necessary, which is needed to support the new graph params). Because it's an InputStream, we don't test the contents for the previously required GRAPH or WITH value. This potentially could cause confusion, since VIVO will accept updates without those values, but the triple won't be visible anywhere to the end user. The documentation should be clear that the user must specify the graph for the update to work properly. 

On the upside, this might have some positive performance implications for large datasets.

# Interested parties
@brianjlowe @VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/api/SparqlUpdateApiController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/api/SparqlUpdateApiController.java
@@ -8,6 +8,7 @@ import static javax.servlet.http.HttpServletResponse.SC_INTERNAL_SERVER_ERROR;
 import static javax.servlet.http.HttpServletResponse.SC_OK;
 
 import java.io.IOException;
+import java.io.InputStream;
 
 import javax.servlet.ServletException;
 import javax.servlet.annotation.WebServlet;
@@ -18,6 +19,8 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import static org.apache.jena.riot.web.HttpNames.paramUsingGraphURI;
+import static org.apache.jena.riot.web.HttpNames.paramUsingNamedGraphURI;
 import org.apache.jena.query.Dataset;
 import org.apache.jena.query.ReadWrite;
 import org.apache.jena.update.GraphStore;
@@ -25,6 +28,9 @@ import org.apache.jena.update.GraphStoreFactory;
 import org.apache.jena.update.UpdateAction;
 import org.apache.jena.update.UpdateFactory;
 import org.apache.jena.update.UpdateRequest;
+import org.apache.jena.graph.Node;
+import org.apache.jena.graph.NodeFactory;
+import org.apache.jena.sparql.modify.UsingList;
 
 import edu.cornell.mannlib.vitro.webapp.application.ApplicationUtils;
 import edu.cornell.mannlib.vitro.webapp.auth.permissions.SimplePermission;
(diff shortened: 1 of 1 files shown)


litvinovg (approved):
Looks ok to me. Works as described.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/api/SparqlUpdateApiController.java
```diff
@@ -76,26 +82,47 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp)
...
+			} catch (IOException e) {
+				log.debug("Error parsing POST body.");
+				throw new ParseException("Error parsing POST body.");
 			}
-			throw new ParseException("SPARQL update must specify a GRAPH ( or WITH) URI.");
 		}
 
 		try {
-			return UpdateFactory.create(update);
+			return UpdateFactory.read(usingList, update);
```
chenejac, line 119: I suggest moving the line 87 just before this line 
UsingList usingList = processProtocol(req);
There is a couple of points where an exception may be thrown, and in that case processing protocol is not necessary.

chenejac (approved):
It works as described. I have just added one comment, @gneissone please check that comment. Moreover, I will add a comment in the linked [issue 1841](https://github.com/vivo-project/VIVO/issues/3429) related to the update of the wiki documentation.

PR #253, opened 2022-02-14
Proof of concept for dynapi february sprint

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3617)**: (please link to issue)

# What's new?
* Contains conceptual elements for creating dynamic actions and dynamic REST APIs.
* A few improvements for configuration bean loader.
* Details are in specification.

# Additional Notes:
Proposal: https://docs.google.com/document/d/1vtNIVEYWdBgV11N-wiPk_UNKpiFQ4sKetJ8elJ6xy2E
Specification: https://docs.google.com/document/d/1n7gSf_cSDS5mbTI4HwVGj-2sQd5if0177Nrdtri9BtM
Sprint wiki: https://wiki.lyrasis.org/display/VIVO/Sprint+-+Dynamic+API+based+on+an+ontology

# Interested parties
@VIVO-project/vivo-committers


api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/RESTEndpoint.java
```diff
@@ -0,0 +1,63 @@
...
+@WebServlet(name = "RESTEndpoint", urlPatterns = { "/api/rest/*" })
+public class RESTEndpoint extends VitroHttpServlet  {
+
+	private static final long serialVersionUID = 1L;
+ 	private static final Log log = LogFactory.getLog(RESTEndpoint.class);
+	private ResourcePool resourcePool = ResourcePool.getInstance();
+ 	
+	
+	@Override
+  public void service(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
```
wwelling, line 24: As a note, we may want to have consistent formatting. Not sure what style VIVO follows more.

wwelling (approved):
I think this is an acceptable starting point. Appears to follow Vitro coding style.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/ActionPool.java
```diff
@@ -0,0 +1,99 @@
...
+		}
+		actions.clear();
+		loadActions();
+	}
+
+	public void init(ServletContext ctx) {
+		this.ctx = ctx;
+		modelAccess = ModelAccess.on(ctx);
+		dynamicAPIModel = modelAccess.getOntModel(FULL_UNION);
+		loader = new ConfigurationBeanLoader(	dynamicAPIModel, ctx);
```
wwelling, line 78: Spurious tab before the first argument of ConfigurationBeanLoad constructor call.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/ActionPool.java
```diff
@@ -0,0 +1,99 @@
...
+		if (INSTANCE == null) {
+			INSTANCE = new ActionPool();
+		}
+		return INSTANCE;
+	}
+	
+	public Action getByName(String name) {
+		Action action = actions.get(name);
+		if (action == null) {
+			action = new DefaultAction();
```
wwelling, line 47: May want to discuss the value of a default action. This action as is will not have a `rpc` or `firstStep` assigned as it is not being loaded.

litvinovg: DefaultAction is a null object to provide behavior if no action is found.

wwelling: Why return an action at all if there are none? What is the expected behavior of the DefaultAction?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/ActionPool.java
```diff
@@ -0,0 +1,99 @@
...
+			return;
+		}
+		if (loader == null ) {
+			log.error("Loader is null. Can't reload action pool.");
+			return;
+		}
+		for (Map.Entry<String,Action> entry : actions.entrySet()) {
+			entry.getValue().dereference();
+		}
+		actions.clear();
```
wwelling, line 70: This will clear the map of actions with a period of time in which no actions will be available. Pending the duration of loading and validating actions. I would suggest loading and validating actions before hot swapping them. May require some synchronization.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/OperationData.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletRequest;
+
+public class OperationData {
+
+	private Map<String, String[]> params;
+	private ServletContext context;
+
+	public OperationData(HttpServletRequest request) {
+		params = request.getParameterMap();
```
wwelling, line 14: Pending the request VERB, will we not want to include doing something with the request body? And potentially headers. Or maybe just put the request on the OperationData, but then what purpose does the OperationData have yet. I assume the OperationData will hold additional "params" as it progresses through the action?

brianjlowe (approved):
I generally agree with William's previous comments (e.g. it seems more intuitive to me to have a lack of action rather than a default action that does nothing), but these seem like minor points we can address in time.  This looks like a great starting point for the sprint.

PR #254, opened 2022-02-19
Dynic API Action Pool Tests

wwelling:


litvinovg (approved):

wwelling:
@litvinovg I did some cleanup and ensured tests are run independent. Action pool singleton needs to be _destroyed_ between each test. Also added the thread safety test here. We will have a failing test until it is resolved.

chenejac (approved):
@wwelling thanks for this. It looks ok to me. Hopefully, we will resolve issues to have passing tests soon. Do we need a similar unit test for ResourcePool?

wwelling:
[ResourcePool](https://github.com/vivo-project/Vitro/blob/sprint-dynapi-2022-feb-staging/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/ResourcePool.java) will need some unit tests but currently it does not expose anything to test assertion. With this action pool testing it shows how to prepare the context for integration tests of the endpoints.

litvinovg (approved):
Looks good to me.

PR #267, opened 2022-02-24
SolrQueryOperation

VeljkoMaksimovic:
# What's new?

* Created an **AbstractOperation** class which all operations (SolrQuery, SPARQLQuery, and Templates) would extend, to avoid code duplication of functions common to all operations, such as validation of input.

* Refactored SPARQLQuery and N3Template to extend newly created AbstractOperation

* Created an **SolrOperation** class, and implemented methods for parsing text queries, inserting input values and creating a query to be executed by SearchEngine.

* Created new action, parameter, parameter type, operation, step and rpc individuals within dynamic-api-individuals-testing.n3 to test SolrOperation

# How should this be tested?
* Created Unit Test class with few test scenarios, but more tests are to be added once we specify structure of OperationResult, and how exactly will the data be returned for operations.

# Additional Notes:
More tests are to be added once we implement functionalities for running an entire action.

EDIT:
Not sure why ActioPoolTest is failing


home/src/main/resources/rdf/abox/filegraph/dynamic-api-individuals-testing.n3
```diff
@@ -23,7 +22,44 @@
...
+        dynapi:requiresParameter <https://vivoweb.org/ontology/vitro-dynamic-api/parameter/textSearchParam> ;
+        dynapi:solrQueryText   "q=?searchText" .
+    
+<https://vivoweb.org/ontology/vitro-dynamic-api/parameter/textSearchParam>
+        a                        dynapi:parameter;
+        dynapi:paramName         "searchText";
+        dynapi:hasParameterType  <https://vivoweb.org/ontology/vitro-dynamic-api/parameter/type/XSDstring> ;
+        rdfs:label               "Search text for solr query 1"@en-US . 
+
+<https://vivoweb.org/ontology/vitro-dynamic-api/action/solrTestAction>
```
wwelling, line 42: Adding this action is causing the action pool test to fail. There are now 2 actions in the pool after init and 3 after reload action is loaded.

VeljkoMaksimovic: Thanks for a quick feedback, will fix today

PR #270, opened 2022-02-24
Move stream reset inside loop so that every listener gets the same state

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3647)**

Re-apply  commit that was accidentally reverted during i18n [sprint](https://github.com/vivo-project/Vitro/commit/eb949919b50c58288ee8936915a8136122fa46fe#)

# How should this be tested?
* Reproduce the problem you are fixing
Create two change listeners, second doesn't get data from InputStream as it was already used
Create two change listeners, second InputStream should get the same data without using multi threading 

# Interested parties
@VIVO-project/vivo-committers @roflinn @brianjlowe  @chenejac


chenejac (approved):
We can expect Benjamin Kampe [commit](https://github.com/bkampe/Vitro/commit/fb9a589a92702f6e79c6f8e0e34225ee282a823a#r67380652) to the sprint staging branch, and hope it wouldn't introduce merging conflict with this one.

wwelling (approved):

PR #271, opened 2022-02-24
[3651] Generic pool key - afford Resource Pool key be version and name

wwelling:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues)**: https://github.com/vivo-project/VIVO/issues/3651

# Interested parties

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -96,21 +96,15 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.easymock</groupId>
-            <artifactId>easymock</artifactId>
-            <version>3.5</version>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>4.3.1</version>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.powermock</groupId>
-            <artifactId>powermock-module-junit4</artifactId>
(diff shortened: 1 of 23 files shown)


chenejac (changes requested):
@wwelling I am happy to see that you found a nice solution for the range key issue. I have just two requests:
1. Can you please consider introducing VersionableAbstractPool as an extension of AbstractPool? ResourcePool should extend VersionableAbstractPool, while ActionPool should extend AbstractPool. The method "public C get(K key)" should be override in VersionableAbstractPool, and four methods with name pattern *Version* should be moved from AbstractPool to VersionableAbstractPool. Any sense in organization of the code as described above?
2. There is a lot of commits in the pull requests including not yours commits. Can you please check how to clean it up, or at least double check whether it could introduce some merging problems?

wwelling:
@chenejac was able to create `VersionableAbstractPool`. Had to add protected `getComponents` on `AbstractPool`. Wasn't so bad at all and made for a better implementation of `get` for the `VersionableAbstractPool`. Thanks for the change request.

wwelling:
@chenejac looks like GitHub finally figured out the redundant commits with staging. :)

If and when approved, we can squash merge to reduce the commits here.

chenejac (approved):
Well done

litvinovg (approved):

PR #265, opened 2022-02-23
Parameter scopes computation

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3662)**

- Created rdf classes (interfaces) in ontology
- Created java interfaces to get data for scopes computation

# Interested parties
Tag (@ mention) interested parties or, if unsure, @VIVO-project/vivo-committers

PR #272, opened 2022-02-24
Solr operation

VeljkoMaksimovic:
# What's new?

Changed SolrQuery to adhere to new interfaces.

The Query itself is no longer stored as a JSON, but rather as a set of RDF properties. Changed Ontology to reflect that change.

Wrote new tests, and changed old ones to use Mockito.

# Still to be done:
Test SolrQuery with ElasticSearch (after sprint)
Write more precise tests once we figure out what will be within OperationResult.
Figure out how exactly are facets used within Vitro app.

# Questions:
I would still like to create AbstractOperation class, if for nothing else, than at least to avoid copying "isValid" method to all Operation classes. There might be even more methods/fields later on that will be the same amongst all Operation classes. Only thing I'm not sure about is whether it is better to "group" those common methods within AbstractOperation class, or OperationUtils class?


api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/SolrQuery.java
```diff
@@ -0,0 +1,141 @@
+package edu.cornell.mannlib.vitro.webapp.dynapi.components;
+
+import com.google.common.reflect.TypeToken;
+import com.google.gson.Gson;
```
wwelling, line 4: Instead of bringing in a new dependency to parse JSON, should we use fasterxml ObjectMapper?

```
private final ObjectMapper objectMapper = new ObjectMapper();
...
TypeReference type = new TypeReference<Map<String, Object>>(){};
return objectMapper.readValue(query, type);
```

May require some exception handling but may be more appropriate than another dependency.

api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/SolrQueryTest.java
```diff
@@ -0,0 +1,96 @@
...
+        replay(parameter1);
+        solrQuery.addRequiredParameter(parameter1);
+        HttpServletRequestStub httpReq = new HttpServletRequestStub();
+        httpReq.addParameter("testParameter","testValue");
+        assertTrue(solrQuery.run(new OperationData(httpReq)).hasError());
+        verify(parameter1);
+    }
+
+    @Test
+    @Ignore
```
wwelling, line 87: Is this still intended to be ignored?

api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/SolrQueryTest.java
```diff
@@ -0,0 +1,96 @@
+package edu.cornell.mannlib.vitro.webapp.dynapi;
+
+import static org.easymock.EasyMock.expect;
```
wwelling, line 3: I would like to bring up discussion on whether we should be using easy mock at all. Before work @kaladay did it was only used on a single line of tests where it wasn't even required. I am not sure it is a good testing framework. It also deviates from all other testing strategy in Vitro. The question then is, do we want to begin defining a new testing strategy? If so, should we use easymock? If neither, it is doable to continue using Vitro testing strategies.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/SolrQuery.java
VeljkoMaksimovic: Ok, will do.

api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/SolrQueryTest.java
VeljkoMaksimovic: I'm not that familiar with different testing strategies and different frameworks, so I can't really add to this discussion. Basically tried to emulate tests used in RPCEndpointTest :) Not sure how to mock all of the components SolrOperation uses, such as Application and SearchEngine without a framework for mocking. Should I create instances of those classes manually before each test scenario? Thanks for detailed comments

api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/SolrQueryTest.java
VeljkoMaksimovic: Yes, for now. Not sure how to evaluate returned value from SolrOperation until we decide what exactly will be returned within OperationResult. There are 5 more test scenarios I am planning on writing, but all require assertions on returned value, so I thought to do it once format of that returned value is agreed upon.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/AbstractOperation.java
```diff
@@ -0,0 +1,48 @@
...
+    protected Parameters requiredParams = new Parameters();
+    protected Parameters providedParams = new Parameters();
+
+    @Property(uri = "https://vivoweb.org/ontology/vitro-dynamic-api#requiresParameter")
+    public void addRequiredParameter(Parameter param) {
+        requiredParams.add(param);
+    }
+
+    @Property(uri = "https://vivoweb.org/ontology/vitro-dynamic-api#providesParameter")
+    public void addProvidedParameter(Parameter param) {
```
litvinovg, line 21: Not all operations provide new parameters.
I added this table a few days ago https://docs.google.com/document/d/1n7gSf_cSDS5mbTI4HwVGj-2sQd5if0177Nrdtri9BtM/edit#heading=h.o146rfjwwctn
 There were some fixes to the ontology that aren't merged yet https://github.com/vivo-project/Vitro/pull/265

api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/SolrQueryTest.java
wwelling: I think it would be valuable to at least now the operation runs all the way through without exception for various inputs.

api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/SolrQueryTest.java
wwelling: Yeah, we can discuss. Maybe we continue to use easymock.

wwelling: We have decided to remove the dependency easymock and use mockito. Is much more intuitive and easier to use. Can find examples in new refactored RPCEndpointTest. Please update to use mockito. https://github.com/vivo-project/Vitro/blob/sprint-dynapi-2022-feb-staging/api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/RPCEndpointTest.java

wwelling (approved):
Looks good to me. We may want to move testing n3 files to src/test/resources at some point.

VeljkoMaksimovic:
> Looks good to me. We may want to move testing n3 files to src/test/resources at some point.

I've done that, and also created a generic method for loading test data within ServletContextTest class, since my guess is that a lot of tests are going to want to load different test data

home/src/main/resources/rdf/tbox/filegraph/vitro-dynamic-api.owl
```diff
@@ -283,6 +284,66 @@
         <rdfs:label xml:lang="en-US">solr query text</rdfs:label>
     </owl:DatatypeProperty>
     
+    <!-- https://vivoweb.org/ontology/vitro-dynamic-api#solrField -->
+    
+    <owl:DatatypeProperty rdf:about="https://vivoweb.org/ontology/vitro-dynamic-api#solrField">
+        <rdfs:subPropertyOf rdf:resource="http://www.w3.org/2002/07/owl#topDataProperty"/>
+        <rdf:type rdf:resource="http://www.w3.org/2002/07/owl#FunctionalProperty"/>
```
chenejac, line 291: I think this can't be FunctionalProperty (https://www.w3.org/TR/owl-ref/#FunctionalProperty-def), if it is defined as functionalProperty there can be maximally one value. Similar is for solrFilter, solrFacets, and solrSort.  We might have a problem in validation that rdf/n3 representations of actions are not in accordance with this ontology (if there is more than one solrField data property.

api/src/test/resources/rdf/abox/filegraph/dynamic-api-individuals-solr-test.n3
```diff
@@ -0,0 +1,59 @@
...
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+
+<https://vivoweb.org/ontology/vitro-dynamic-api/solrQuery/genericSolrTextQuery>
+        a                        dynapi:operation, dynapi:solrQuery ;
+        rdfs:label               "Test solr query 1"@en-US ;
+        dynapi:requiresParameter <https://vivoweb.org/ontology/vitro-dynamic-api/parameter/textSearchParam> ;
+        dynapi:solrQueryText   "http" ;
+        dynapi:solrField   "Field1" ;
+        dynapi:solrField   "Field2" ;
```
chenejac, line 15: Usually this is structured as:
dynapi:solrField    "Field1" , "Field2" ;

or

dynapi:solrField    "Field1" , 
                             "Field2" ;

chenejac (changes requested):
@VeljkoMaksimovic I have posted a couple of comments. Please consider whether it might be adopted.

wwelling (approved):

PR #273, opened 2022-02-25
[3622] Add getMethod() and rpcMethodUnsupported(), also updating tests accordingly.

kaladay:
**[VIVO GitHub issue 3622](https://github.com/vivo-project/VIVO/issues/3622)**:

The RPC Endpoint needs to be aware of the correctness of the HTTP Method received.
If the Method is not Implemented, then properly return this.
The facility for doing this is missing and so this PR introduces this needed behavior for returning "501 Not Implemented" as appropriate.

# What's new?
Add `getMethod()` to `OperationData` to facilitate this.
Add `rpcMethodUnsupported()` to `Action` and utilize `rpcMethodUnsupported()` to validate the HTTP Method.

Some `RPCEndpointTest` tests are updated to accomodate this.
Remove some unecessary steps from `RPCEndpointTest`, as the `@Mock` already handles mock instantiation.

# How should this be tested?
* This mostly adds new methods, so the existing RPC Endpoints should be tested to continue to work in the same way they did before this PR.

# Additional Notes:
* This does not solve the aforementioned issue, this PR instead improves the RPC Endpoint implementation.
* There appears to be more work needed to properly verify the Methods than simply adding these methods.

# Interested parties

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/Action.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/Action.java
@@ -1,5 +1,11 @@
 package edu.cornell.mannlib.vitro.webapp.dynapi.components;
 
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
 import javax.servlet.http.HttpServletResponse;
 
 import org.apache.commons.logging.Log;
@@ -7,12 +13,6 @@ import org.apache.commons.logging.LogFactory;
 
 import edu.cornell.mannlib.vitro.webapp.dynapi.OperationData;
 import edu.cornell.mannlib.vitro.webapp.utils.configuration.Property;
-
-import java.util.Collections;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.stream.Collectors;
 
 public class Action implements RunnableComponent, Poolable, Operation, Link {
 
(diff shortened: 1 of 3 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/Action.java
```diff
@@ -122,4 +122,12 @@ public boolean isRoot() {
     return true;
   }
 
+	private boolean rpcMethodUnsupported(String method) {
+		if (rpc == null || rpc.getHttpMethod() == null || rpc.getHttpMethod().getName() == null) {
+			return false;
```
kaladay, line 127: I'm looking at my own PR and thinking maybe this should be returning true here instead of false.

wwelling: It appears so. To avoid the negative logic confusion, I suggest the following:

```
	public OperationResult run(OperationData input) {
		if (firstStep != null && rpcMethodSupported(input.getMethod())) {
			return firstStep.run(input);			
		}
		return new OperationResult(HttpServletResponse.SC_NOT_IMPLEMENTED);

	}
	
	...
	
	private boolean rpcMethodSupported(String method) {
		return rpc != null &&
			rpc.getHttpMethod() != null &&
			rpc.getHttpMethod().getName() != null &&
			rpc.getHttpMethod().getName().equalsIgnoreCase(method);
	}

```

wwelling: Or at least correct such as

```
	public OperationResult run(OperationData input) {
		if (firstStep == null || rpcMethodUnsupported(input.getMethod())) {
			return new OperationResult(HttpServletResponse.SC_NOT_IMPLEMENTED);
		}
		return firstStep.run(input);
	}
	
	...
	
	private boolean rpcMethodUnsupported(String method) {
		return rpc == null ||
			rpc.getHttpMethod() == null ||
			rpc.getHttpMethod().getName() == null ||
			!rpc.getHttpMethod().getName().equalsIgnoreCase(method);
	}
```

kaladay:
Closing in favor of moving the HTTP Method check directly into the RPC Endpoint Servlet.

PR #276, opened 2022-02-25
Properly handle unimplemented Methods in RPC and replace testing framework.

kaladay:
**[VIVO GitHub issue 3622](https://github.com/vivo-project/VIVO/issues/3622)**:

* Change RPC Endpoint to return 501 for some REST methods that are not supported.
* Fully replaces EasyMock (and PowerMock) with Mockito.

# What's new?
Uses `request.getPathInfo()` to parse the action name from the URI.

The RPC endpoints should return 501 for the GET, PUT, and DELETE methods.

The tests are updated to handle this.
Using EasyMock continued to be enough problems that I feel that it is necessary to replace EasyMock (and the recently added PowerMock) with Mockito.

Mockito allows for much cleaner test writing and provides a more stable API.

The other projects depending on EasyMock either use it in very few ways or are using it when they should instead be using JUnit (specifically the JUnit assertions).

# Interested parties

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/RPCEndpoint.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/RPCEndpoint.java
@@ -12,47 +12,41 @@ import edu.cornell.mannlib.vitro.webapp.dynapi.components.Action;
 import edu.cornell.mannlib.vitro.webapp.dynapi.components.OperationResult;
 
 @WebServlet(name = "RPCEndpoint", urlPatterns = { "/api/rpc/*" })
-public class RPCEndpoint extends VitroHttpServlet  {
+public class RPCEndpoint extends VitroHttpServlet {
 
 	private static final long serialVersionUID = 1L;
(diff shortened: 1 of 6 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/RPCEndpoint.java
```diff
@@ -12,47 +12,41 @@
...
+	public void doGet(HttpServletRequest request, HttpServletResponse response) {
+		OperationResult.notImplemented().prepareResponse(response);
 	}
-	
+
 	@Override
-	public void doPost( HttpServletRequest request, HttpServletResponse response ) {
-		process(request, response);
+	public void doPost(HttpServletRequest request, HttpServletResponse response) {
+		String requestURL = request.getRequestURI();
```
wwelling, line 28: We should use request.getPathInfo() and verify it has at least one path. Else should return bad request.

wwelling (approved):

litvinovg (approved):
Looks good to me.

PR #278, opened 2022-03-01
[3678] Implement integration testing for RPC Endpoints.

kaladay:
**[VIVO GitHub issue 3678](https://github.com/vivo-project/VIVO/issues/3678)**:

Add integration testing for the RPC Endpoints.

# What's new?
The goal here is to mock as little as possible and to test on what goes and and what comes out of the endpoint.

Because the data is auto-loaded (via `ConfigurationBeanLoader` in `ServletContextTest`) from the RDF data, the QueryExecutionFactory itself is mocked.
This operates on the assumption that the RDF is executing a SPARQL query.
The SPARQL query should not run for the purposes of this test and is instead mocked.
Every other aspect of this testing process avoids mocking as much as possible.

Avoid `@RunWith(MockitoJUnitRunner.class)` and the mocking annotations so that paremeterized tests can be used.
(Junit5 would be a boon here, given that Junit4 doesn't support multiple simultaneously runners.)

A significant amount of code is reduced by utilizing parameterized testing.
This has a downside of requiring certain framework to be in place (and Junit4 has notable limitations in this regard).
The testing data is defined within the `requests()` method and is written using a custom styling that makes the parameters easier to understand and modify.

The GET, PUT, and DELETE endpoints should always return 501 (Not Implemented).
By utilizing the parameter testing, these testing process ensures that even if the Request paremters are different, the 501 is always returned.

The POST endpoint handles the specific endpoints.
I would note that a 500 is returned for validation failures, but wouldn't these be better suited to being 400 (Bad Request)s?

Found and fixed a problem in `dynamic-api-individuals-testing.n3` where `testParam2` most likely should be referring to 'email' rather than 'limit'.

# Interested parties


wwelling (approved):

home/src/main/resources/rdf/abox/filegraph/dynamic-api-individuals-testing.n3
```diff
@@ -36,11 +36,11 @@
 
 <https://vivoweb.org/ontology/vitro-dynamic-api/parameter/testParam2>
         a                                   dynapi:parameter ;
-        dynapi:paramName                    "limit" ;
+        dynapi:paramName                    "email" ;
```
chenejac, line 39: Thanks for this. This was my copy-paste error, it is better to be email.

chenejac (approved):

wwelling (approved):

chenejac (changes requested):

chenejac (approved):

PR #279, opened 2022-03-03
[3667] Validate REST method and update unit tests

wwelling:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues)**:

Validates request method to the RPC method for both REST requests and custom REST action requests. Updated REST endpoint unit tests accordingly.

# Interested parties

### DIFF
--- a/api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/RESTEndpointTest.java
+++ b/api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/RESTEndpointTest.java
@@ -3,12 +3,15 @@ package edu.cornell.mannlib.vitro.webapp.dynapi;
 import static edu.cornell.mannlib.vitro.webapp.dynapi.request.RequestPath.REST_BASE_PATH;
 import static org.apache.commons.lang3.StringUtils.EMPTY;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.mockStatic;
 import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
 import java.util.Map;
 
 import javax.servlet.ServletContext;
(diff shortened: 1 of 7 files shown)


api/src/test/java/edu/cornell/mannlib/vitro/webapp/dynapi/RESTEndpointTest.java
```diff
@@ -19,31 +22,34 @@
...
 
 import edu.cornell.mannlib.vitro.webapp.dynapi.components.Action;
+import edu.cornell.mannlib.vitro.webapp.dynapi.components.HTTPMethod;
 import edu.cornell.mannlib.vitro.webapp.dynapi.components.OperationResult;
+import edu.cornell.mannlib.vitro.webapp.dynapi.components.RPC;
 import edu.cornell.mannlib.vitro.webapp.dynapi.components.Resource;
 import edu.cornell.mannlib.vitro.webapp.dynapi.components.ResourceKey;
 
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(Parameterized.class)
```
kaladay, line 39: This change here potentially introduces a problem.
The `MockitoJUnitRunner.class` is what is supposed to process the annotations such as `@Mock` and `@Spy`.
I'm surprised there are no errors, maybe the mocks aren't needed?

For every `@Mock` you should in instead have a `Mockito.mock(XXX);` for creating a mock manually.
The spy is `Mockito.spy(XXX)`.

If I am wrong, then that is great and I can simplify my integration tests to use the annotations rather than directly calling the `Mockito.mock()` and `Mockito.spy()` methods.

I will investigate this some to determine if this is indeed a problem or is not the problem that I am thinking it is.

kaladay: Okay I see.
It seems that you introduced this:
```
MockitoAnnotations.openMocks(this);
```

Which processes the annotations, allowing for those annotations to be used when using a different runner.
I wasn't aware of this.
I can use this to improve the tests that I wrote.

wwelling: They error without that openMocks call.

kaladay (approved):

chenejac (approved):

PR #288, opened 2022-04-23
 Prevent copying pom file to home directory

litvinovg:
Changes artifact type from pom to tar.gz to prevent copying pom files ( vitro-languages-home-core-1.12.3-SNAPSHOT.pom) to home directory.

**How to test**
Install Vitro from sources before applying, notice that file is added to home directory.
Clean home directory
Apply this PR
Install Vitro from sources, verify that file is not added to home directory


chenejac (approved):
Works for me.

gneissone:
Vitro built and ran. No extra pom file found its way to the home directory.

gneissone (approved):

PR #292, opened 2022-05-27
VIVO-3713 Avoid excessive use of resources when ABox firsttime data are changed

brianjlowe:
**[VIVO-3713](https://github.com/vivo-project/VIVO/issues/3713)**:

No functional changes.  Modifies the "firsttime" update code that now runs on each startup to remove a diff against the entire graph being updated, which may cause a very large graph (kb-2) to be attempted to be read into memory during an upgrade to 1.12.x or if the contents of the ABox firsttime files change.

# What's new?
- Refactors the two checkUiChangesOverlapWithFileChanges() methods into one utility method removeChangesThatConflictWithUIEdits() located in RDFFilesLoader.
- Changes userModel.difference() to scopedUserModel.difference(), where scopedUserModel contains only triples with subject/predicate pairs that occur in the set of changed firsttime triples (i.e., only those triples that are actual candidates for being updated, not the entire model).
- Improves StmtIterator iter2 by retrieving only triples with the desired subject/predicate pair instead of iterating over the entire in-memory model.
- Ignores isIsomorphicWith() failures that are only due to blank node data.  Previously, these were causing the update code to run in certain cases but then produce no changes because the updater ignored blank nodes anyway.
- Adds unit tests for removeChangesThatConflictWithUIEdits

# How should this be tested?
If the unit tests in RDFFilesLoader pass, it confirms that the firsttime update behavior is as desired under the new implementation of checkUiChangesOverlapWithFileChanges().  If additional manual testing is desired, it would be the same as for https://github.com/vivo-project/VIVO/issues/3505 .

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/servlet/setup/RDFFilesLoader.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/servlet/setup/RDFFilesLoader.java
@@ -224,7 +224,7 @@ public class RDFFilesLoader {
 	 * @param userModel current state in the system (user/UI-model)
 	 * @param changesModel the changes between firsttime-files and firsttime-backup
 	 */
-	public static void checkUiChangesOverlapWithFileChanges(Model baseModel,
+	public static void removeChangesThatConflictWithUIEdits(Model baseModel,
 	        Model userModel, Model changesModel) {
 	    log.debug("Check if subtractions from backup-firsttime model to"
 	            + " current state of firsttime-files were changed in user-model"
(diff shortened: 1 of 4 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/servlet/setup/RDFFilesLoader.java
```diff
@@ -207,5 +214,133 @@ private static String locateHomeDirectory() {
...
+	/**
+	 * Check if the user model (UI) changes conflict with the changes made to
+	 * the firsttime.  If there is conflict, the user model UI value will be
+	 * left unchanged.
+	 * 
+	 * @param baseModel firsttime backup model
+	 * @param userModel current state in the system (user/UI-model)
+	 * @param changesModel the changes between firsttime-files and firsttime-backup
+	 */
+	public static void checkUiChangesOverlapWithFileChanges(Model baseModel,
```
litvinovg, line 227: If I understood, it excludes some triples from changesModel. 
Maybe it would be better to rename it to something like excludeChangesMadeInUi ?

litvinovg (approved):
Looks good to me.

brianjlowe:
Georgy pointed out that setPortalUriOnFirstTime() is also problematic because it only renames the first Portal node it finds, and there may be multiple Portal nodes (see https://github.com/vivo-project/VIVO-languages/pull/108).  We could remove those additional nodes entirely from the language files, but someday maybe someone would want to set a language-specific site name.  It would be safer to update the portal renaming to use the same URI for all nodes it finds.  Will update this PR.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/servlet/setup/RDFFilesLoader.java
brianjlowe: changed to removeChangesThatConflictWithUIEdits()

litvinovg (approved):

chenejac (approved):

PR #293, opened 2022-05-30
Fix class uri for validators

litvinovg:
There are Validator and validator classes listed the in ontology, this PR leaves only validator class.


chenejac (approved):

brianjlowe:
It seems like this should go the other way?  The typical convention if not using numeric URIs is UpperCamelCase for classes and lowerCamelCase for properties.

chenejac (approved):

PR #295, opened 2022-06-01
Individual list calculation fix

litvinovg:
If individual list is big enough (4584 for example), last page (153) would be missing on all pages other than 113, because of rounding toward negative infinity 4584/30 = 152.8 , after rounding it would be 152. 

The problem starts to appear from 1231 individuals 40 pages (30 individuals on each ) + more button represents 41 page( 30 more )

### How to test:
Create 4584 individuals, open individuals list, go to page greater than 113, observe that page 153 is visible.
# Interested parties


gneissone (approved):
Confirmed issue. Fix tested working.

chenejac (approved):

PR #296, opened 2022-06-02
fix: process string and untyped literals differently

litvinovg:
Fixes for https://github.com/vivo-project/VIVO/issues/3717

Applied on https://vivo.tib.eu/vivo113rc

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
@@ -307,7 +307,7 @@ public class MultiValueEditSubmission {
 							}
 							// if the language is set in the given Literal, this language-tag should be used and remain the same
 							// for example when you edit an label with an langauge-tag (no matter which language is selected globally)
-							if (ifLangSetInFirstLiteral(editConfig, getLabelLanguage) )
+							if (getLabelLanguage && isLangSetInFirstLiteral(editConfig) )
 							{
 								rangeLang = editConfig.getLiteralsInScope().get(LABEL).get(0).getLanguage();
 							} else { // if the literal has no langauge-tag, use the language which is globally selected
@@ -334,10 +334,7 @@ public class MultiValueEditSubmission {
         }
     }
 
-    private boolean ifLangSetInFirstLiteral(EditConfigurationVTwo editConfig, Boolean getLabelLanguage) {
-        if (!getLabelLanguage) {
-            return false;
-        }
+    private boolean isLangSetInFirstLiteral(EditConfigurationVTwo editConfig) {
         Map<String, List<Literal>> literalsInScope = editConfig.getLiteralsInScope();
         if (!literalsInScope.containsKey(LABEL)) {
             return false;
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
@@ -30,6 +30,8 @@ import edu.cornell.mannlib.vitro.webapp.i18n.I18n;
 
 public class MultiValueEditSubmission {
 
+    private static final String LABEL = "label";
+
     String editKey;
 
     private Map<String,List<Literal>> literalsFromForm ;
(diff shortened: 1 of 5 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
```diff
@@ -172,20 +174,14 @@ public Literal createLiteral_ORIG(String value, String datatypeUri, String lang)
...
-			// UQAM Original code contained tow-dots ':' in place of '#'
-//            if( "http://www.w3.org/2001/XMLSchema:anyURI".equals(datatypeUri) ){
 			if( XSD.anyURI.getURI().equals(datatypeUri) ){
-//				try {
-//					return literalCreationModel.createTypedLiteral( URLEncoder.encode(value, "UTF8"), datatypeUri);
 					return literalCreationModel.createTypedLiteral( value, datatypeUri);
-//				} catch (UnsupportedEncodingException e) {
-//					log.error(e, e);
-//				}
-			} else if ( XSD.xstring.getURI().equals(datatypeUri) || RDF.dtLangString.getURI().equals(datatypeUri) ){
```
wwelling, line 184: Hi @litvinovg, could you help me understand why `XSD.xstring.getURI()` is no longer a valid condition here and below?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
```diff
@@ -172,20 +174,14 @@ public Literal createLiteral_ORIG(String value, String datatypeUri, String lang)
...
-				if( lang != null && lang.length() > 0 )	return ResourceFactory.createLangLiteral(value, lang);
+			} else if ( RDF.dtLangString.getURI().equals(datatypeUri) ){
+				if( lang != null && lang.length() > 0 )	{
```
wwelling, line 180: As we are here, this would be better using `StringUtils` as done below. `StringUtils.isNotEmpty(lang)`

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
```diff
@@ -336,6 +333,23 @@ public void addLiteralToForm(EditConfigurationVTwo editConfig, FieldVTwo field,
...
+        List<Literal> labelLiterals = literalsInScope.get(LABEL);
+        if (labelLiterals.size() == 0) {
+            return false;
+        }
+        Literal literal = labelLiterals.get(0);
+        if (literal == null){
+            return false;
+        }
+        
+        return !StringUtils.isBlank(literal.getLanguage());
```
wwelling, line 351: `StringUtils.isBlank` will return true for and whitespace character. That will likely cause an issue the same as an empty string. Additionally, I would recommend the negation StringUtils method `isNotEmpty`

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/MultiValueEditSubmission.java
```diff
@@ -336,6 +333,23 @@ public void addLiteralToForm(EditConfigurationVTwo editConfig, FieldVTwo field,
...
         }
     }
+
+    private boolean isLangSetInFirstLiteral(EditConfigurationVTwo editConfig) {
+        Map<String, List<Literal>> literalsInScope = editConfig.getLiteralsInScope();
+        if (!literalsInScope.containsKey(LABEL)) {
+            return false;
+        }
+        List<Literal> labelLiterals = literalsInScope.get(LABEL);
+        if (labelLiterals.size() == 0) {
```
wwelling, line 343: Aside: I am not a fan of all this uncertainty in code. Would be to be certain an empty list was not added to the map, would be nice to be certain a null was not added to said list, etc.

With that said, shouldn't we check to see if the list in the map is not null?

wwelling (dismissed):
Code review looks good.

gneissone (changes requested):
I'm experiencing different behaviors when editing via the front end editor form vs the 'Edit this individual' button

Initial value after creating new faculty member
Screen Shot 2022-06-30 at 5 27 16 PM

Edit via front end....
Screen Shot 2022-06-30 at 5 30 48 PM

Keeps language tags. I believe this is the intended behavior. 
Screen Shot 2022-06-30 at 5 31 30 PM

Edit via 'edit this individual'...
Screen Shot 2022-06-30 at 5 28 08 PM

Language tags gone
Screen Shot 2022-06-30 at 5 28 21 PM

Can similar fixes be applied to EntityRetryController.java so it behaves in the same way?

gneissone:
Okay, at the very least we can document the incongruity. Perhaps for now we can add a disclaimer to the backend editing form alerting the user that the form will remove language tags (only if multilanguage is enabled?? or always??)

gneissone (approved):
Works as expected. To mitigate the issue described in my previous review I would like to see https://github.com/vivo-project/Vitro/pull/313 merged.

chenejac (approved):

PR #299, opened 2022-06-07
Create initial Login controller with tests

VeljkoMaksimovic:
Created initial Login controller that uses already existing Authorization module for authenticating users, and later saving them to a session


wwelling (changes requested):
Could we assert the existence on the session cookie on successful login response? There are also a few formatting fixes that can be made with missing spaces between method and starting code block bracket. Otherwise, this looks good.

wwelling:
@VeljkoMaksimovic looks like I asked for something that existing testing strategy doesn't afford, https://github.com/vivo-project/Vitro/blob/4ed88b81e3738fccbf3beef677bc165bcaa37179/api/src/test/java/stubs/javax/servlet/http/HttpServletResponseStub.java#L250. Honestly, I think we should avoid using these existing stubs and mock the response the same as the request and verify appropriate method calls are made on the response.

However, the concern I have is how the browser will provide session in subsequent requests after log in. I was assuming this would be done by a session cookie on the response to login. Yet it does not appear the login process tested calls add cookie.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/authentication/LoginController.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.jena.base.Sys;
+
+import javax.servlet.ServletException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import java.io.IOException;
+
+@WebServlet(name = "Login", urlPatterns = {"/rest_login"})
```
wwelling, line 18: Can we follow the existing dynamic API request paths?

https://github.com/vivo-project/Vitro/blob/sprint-dynapi-2022-feb-staging/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/RESTEndpoint.java#L27

I would suggest `/api/login` and possibly create constant such as https://github.com/vivo-project/Vitro/blob/sprint-dynapi-2022-feb-staging/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/request/ApiRequestPath.java#L15

```
API_SERVLET_PATH + "/login";
```

wwelling:
@VeljkoMaksimovic tested the login via postman and it does indeed create the session cookie. It does it via Set-Cookie header. Can you add assertion in the login test that the response does indeed have the Set-Cookie header? And please consider switching to a mock response opposed to using existing Vitro stubs.

image

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/authentication/LoginController.java
```diff
@@ -0,0 +1,73 @@
...
+            return;
+        }
+
+        request.getSession().setAttribute("logged_in_user", user);
+        response.setStatus(200);
+    }
+
+    private void writeResponseError(HttpServletResponse response, String message) {
+        try {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST, message);
```
wwelling, line 68: We should provide 401 for when authentication fails but the request is valid. https://www.rfc-editor.org/rfc/rfc7235#section-3.1

If missing username or password parameters 400 is valid.

wwelling:
Sorry I missed a few points on first review.

VeljkoMaksimovic:
> @VeljkoMaksimovic tested the login via postman and it does indeed create the session cookie. It does it via Set-Cookie header. Can you add assertion in the login test that the response does indeed have the Set-Cookie header? And please consider switching to a mock response opposed to using existing Vitro stubs.
> 
> image
> 
> image

Not sure I can check whether response has a Set-Cookie header since response is a mock. I thought to verify whether response.setHeader(eq("Set-Cookie",any(String.class)) was called exactly once, but that can't be done inside the unit test since session header is not added inside the LoginController.doPost or .process() methods, but later in the chain of executed functions. All I could check is whether the ".getSession" and .setAttribute("logged_in_user") method were called

VeljkoMaksimovic:
> Sorry I missed a few points on first review.

no problem

wwelling:
Will need an integration test for the endpoint.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/authentication/LoginController.java
```diff
@@ -15,7 +14,9 @@
 
 import java.io.IOException;
 
-@WebServlet(name = "Login", urlPatterns = {"/rest_login"})
+import static edu.cornell.mannlib.vitro.webapp.dynapi.request.ApiRequestPath.API_SERVLET_LOGIN;
+
+@WebServlet(name = "Login", urlPatterns = {API_SERVLET_LOGIN})
 public class LoginController extends VitroHttpServlet {
 
     private static final Log log = LogFactory.getLog(RESTEndpoint.class);
```
wwelling, line 22: Logger is not used. Is also for wrong class.

wwelling (approved):
Please remove logger or update its class.

I think it may be too much to ask for integration tests for the existing Vitro authenticator. That being the only way to really assert client will be logged in, we will have to settle for manual testing until strategy for login controller integration tests is derived.

SoulsMark (approved):

PR #305, opened 2022-06-10
N3 operation retractions

VeljkoMaksimovic:
Extended N3TemplateOperation (both definition and implementation) to also support retraction of triplets.

Question
How to delete multiple entries with one n3 triplet? For example, delete all statements that start with <http://Joe> <http://Has>

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
@@ -78,7 +78,8 @@ public class N3Template extends Operation implements Template {
 			return new OperationResult(500);
 		}
 
-		List<Model> additionModels, retractionModels;
+		List<Model> additionModels;
+		List<Model> retractionModels;
 		try {
 			additionModels = ProcessRdfForm.parseN3ToRDF(
 					Arrays.asList(substitutedN3AdditionsTemplate),
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
@@ -68,11 +68,11 @@ public class N3Template extends Operation implements Template {
 			return new OperationResult(500);
 		}
 
-		String substitutedN3AdditionsTemplate = "";
-		String substitutedN3RetractionsTemplate = "";
+		String substitutedN3AdditionsTemplate;
+		String substitutedN3RetractionsTemplate;
 		try {
-			n3TextAdditions = substitutedN3AdditionsTemplate=insertParameters(input, n3TextAdditions);
-			n3TextRetractions = substitutedN3RetractionsTemplate=insertParameters(input, n3TextRetractions);
+			substitutedN3AdditionsTemplate = insertParameters(input, n3TextAdditions);
+			substitutedN3RetractionsTemplate = insertParameters(input, n3TextRetractions);
 		}catch (InputMismatchException e){
 			log.error(e);
 			return new OperationResult(500);
@@ -80,8 +80,13 @@ public class N3Template extends Operation implements Template {
 
 		List<Model> additionModels, retractionModels;
 		try {
-			additionModels = ProcessRdfForm.parseN3ToRDF(Arrays.asList(substitutedN3AdditionsTemplate), ProcessRdfForm.N3ParseType.REQUIRED);
-			retractionModels = ProcessRdfForm.parseN3ToRDF(Arrays.asList(substitutedN3RetractionsTemplate), ProcessRdfForm.N3ParseType.REQUIRED);
+			additionModels = ProcessRdfForm.parseN3ToRDF(
+					Arrays.asList(substitutedN3AdditionsTemplate),
+					ProcessRdfForm.N3ParseType.REQUIRED);
+
+			retractionModels = ProcessRdfForm.parseN3ToRDF(
+					Arrays.asList(substitutedN3RetractionsTemplate),
+					ProcessRdfForm.N3ParseType.REQUIRED);
 		} catch (Exception e) {
 			log.error("Error while trying to parse N3Template string and create a Jena rdf Model", e);
 			return new OperationResult(500);
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
@@ -19,8 +19,8 @@ public class N3Template extends Operation implements Template {
 
 
 	private Parameters requiredParams = new Parameters();
-	private String n3TextAdditions;
-	private String n3TextRetractions;
+	private String n3TextAdditions = "";
+	private String n3TextRetractions = "";
 	private ModelComponent templateModel;
 
     // region @Property Setters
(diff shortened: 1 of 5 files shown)


chenejac:
> For now, both n3TextAdditions and n3TextRetraction property cardinalities are set to min:0, max:1. That allows to have n3Template that has no addition statements, nor retraction statements, which should not be valid. Not sure how to set "collective cardinality" on those two properties to be min:1

Probably this is a task for me to be resolved by definition of a SHACL rule.

litvinovg:
I think we need integration tests to check that action with n3 template is converted into java action with configurationBeanLoader and instantiated action with n3 template can process operation data.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -19,7 +19,8 @@ public class N3Template extends Operation implements Template {
 
 
 	private Parameters requiredParams = new Parameters();
-	private String n3Text;
+	private String n3TextAdditions;
```
litvinovg, line 22: Maybe initialize empty string to get rid of null checks?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -61,31 +69,35 @@ public OperationResult run(OperationData input) {
 			return new OperationResult(500);
 		}
 
-		String substitutedN3Template;
+		String substitutedN3AdditionsTemplate = null, substitutedN3RetractionsTemplate = null;
 		try {
-			substitutedN3Template=insertParameters(input);
+			if (this.n3TextAdditions != null) substitutedN3AdditionsTemplate=insertParameters(input, this.n3TextAdditions);
```
litvinovg, line 74: I would suggest to always use curly braces

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -61,31 +69,35 @@ public OperationResult run(OperationData input) {
...
+		List<Model> additionModels = new ArrayList<>(), retractionModels = new ArrayList<>();
 		try {
-			additionModels = ProcessRdfForm.parseN3ToRDF(Arrays.asList(substitutedN3Template), ProcessRdfForm.N3ParseType.REQUIRED);
+			if (substitutedN3AdditionsTemplate != null)
+				additionModels = ProcessRdfForm.parseN3ToRDF(Arrays.asList(substitutedN3AdditionsTemplate), ProcessRdfForm.N3ParseType.REQUIRED);
+			if (substitutedN3RetractionsTemplate != null)
+				retractionModels = ProcessRdfForm.parseN3ToRDF(Arrays.asList(substitutedN3RetractionsTemplate), ProcessRdfForm.N3ParseType.REQUIRED);
 		} catch (Exception e) {
 			log.error("Error while trying to parse N3Template string and create a Jena rdf Model");
 			log.error(e);
```
litvinovg, line 89: I would suggest log.error(e,e);

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -61,31 +69,35 @@ public OperationResult run(OperationData input) {
...
 			return new OperationResult(500);
 		}
 
-		List<Model> additionModels;
+		List<Model> additionModels = new ArrayList<>(), retractionModels = new ArrayList<>();
 		try {
-			additionModels = ProcessRdfForm.parseN3ToRDF(Arrays.asList(substitutedN3Template), ProcessRdfForm.N3ParseType.REQUIRED);
+			if (substitutedN3AdditionsTemplate != null)
+				additionModels = ProcessRdfForm.parseN3ToRDF(Arrays.asList(substitutedN3AdditionsTemplate), ProcessRdfForm.N3ParseType.REQUIRED);
```
litvinovg, line 84: curly braces

litvinovg (changes requested):
It makes sense to rename n3Template to N3Template in ontology as you fixing this.

brianjlowe:
Agree with all of Georgy's requested changes.  In addition, watch line length.  I usually try to not to exceed 80 characters (which may be more conservative than necessary), but as it currently some of the lines needing curly braces overflow the div and are hidden when using Github's Show File view.  This suggests that they are simply too long.

When Georgy's satisfied with the changes, I can add my second approval.  While I don't fully understand the context or use cases, in isolation the approach of adding a retractions template and processing it via the same mechanism as additions seems logical enough.

VeljkoMaksimovic:
> Agree with all of Georgy's requested changes. In addition, watch line length. I usually try to not to exceed 80 characters (which may be more conservative than necessary), but currently some of the lines needing curly braces overflow the div and are hidden when using Github's Show File view. This suggests that they are simply too long.
> 
> When Georgy's satisfied with the changes, I can add my second approval. While I don't fully understand the context or use cases, in isolation the approach of adding a retractions template and processing it via the same mechanism as additions seems logical enough.

Ok, thanks for the suggestion, will do later in the day

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -61,31 +68,33 @@ public OperationResult run(OperationData input) {
 			return new OperationResult(500);
 		}
 
-		String substitutedN3Template;
+		String substitutedN3AdditionsTemplate = "";
+		String substitutedN3RetractionsTemplate = "";
 		try {
-			substitutedN3Template=insertParameters(input);
+			n3TextAdditions = substitutedN3AdditionsTemplate=insertParameters(input, n3TextAdditions);
```
litvinovg, line 74: formatting + n3TextAdditions shouldn't be changed during execution.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -61,31 +68,33 @@ public OperationResult run(OperationData input) {
...
+			n3TextRetractions = substitutedN3RetractionsTemplate=insertParameters(input, n3TextRetractions);
```
litvinovg, line 75: formatting + n3TextRetractions shouldn't be changed during execution.

litvinovg (changes requested):

litvinovg:
Please add test to verify that after one run of n3template with input data second run with different input data produces different results.

VeljkoMaksimovic:
> Please add test to verify that after one run of n3template with input data second run with different input data produces different results.

Ok, I added one extra test that does that.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -61,31 +68,38 @@ public OperationResult run(OperationData input) {
...
-			substitutedN3Template=insertParameters(input);
+			substitutedN3AdditionsTemplate = insertParameters(input, n3TextAdditions);
+			substitutedN3RetractionsTemplate = insertParameters(input, n3TextRetractions);
 		}catch (InputMismatchException e){
 			log.error(e);
 			return new OperationResult(500);
 		}
 
-		List<Model> additionModels;
+		List<Model> additionModels, retractionModels;
```
litvinovg, line 81: Please declare variables on separate lines

litvinovg (approved):
Looks good to me.

brianjlowe (approved):

PR #310, opened 2022-06-15
N3 operation retractions

VeljkoMaksimovic:
@chenejac noticed that n3 templates dont work with complex objects as input, which can be the case when output of one operation, like SPARQL is input to N3 template operation.

This PR is not yet done, and needs more changes

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
@@ -12,6 +12,7 @@ import org.apache.jena.rdf.model.Literal;
 import org.apache.jena.rdf.model.Model;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 public class N3Template extends Operation implements Template {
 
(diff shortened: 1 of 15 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -104,20 +109,58 @@ private String insertParameters(OperationData input, String n3Text) throws Input
...
+//		Map<String, List<String>> parametersToUris = requiredParams.getUrisMap(input);
+//
+//		gen.subInMultiUris(parametersToUris, n3WithParameters);
+//		//endregion
+//
+//		//region Substitute other (literal) variables
+//		Map<String, List<Literal>> parametersToLiterals = requiredParams.getLiteralsMap(input);
+//
+//		gen.subInMultiLiterals(parametersToLiterals, n3WithParameters);
+//		//endregion
```
chenejac, line 163: I suggest removing those lines, if @VeljkoMaksimovic don't think it might be useful again in the future.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -6,15 +6,20 @@
...
 import org.apache.jena.rdf.model.Literal;
 import org.apache.jena.rdf.model.Model;
+import org.apache.jena.rdf.model.ResourceFactory;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 public class N3Template extends Operation implements Template {
 
+	private static final String ANY_URI="http://www.w3.org/2001/XMLSchema#anyURI";
```
chenejac, line 21: I believe we have a constant for ANY_URI defined a couple of times, probably we should change access modifier (to public), and defined that at one place.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
```diff
@@ -129,7 +172,12 @@ private String insertParameters(OperationData input, String n3Text) throws Input
 		return n3WithParameters.get(0);
 	}
 
-
+	private List<String> getVariables(String n3Text){
+    	return Arrays.stream(n3Text.split("\\s+"))
```
chenejac, line 176: please fix indentation here

chenejac (changes requested):

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
VeljkoMaksimovic: I would leave them here for now, just until we decide whether the right place for substituting variables with concrete values is inside N3Template class (where they are currently) or inside Parameters class (where these commented out lines are). @litvinovg

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/N3Template.java
VeljkoMaksimovic: Just looked at it, and it seams we have this constant only in N3Templates and Parameters. Again, once we decide where variable substitution belongs, we can leave this constant either here, or inside Parameters class

chenejac:
resolved within https://github.com/vivo-project/Vitro/pull/319

PR #314, opened 2022-07-04
Workaround for creating integer literals 

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3728)**

Fixes issue related to creating integer literals in web interface by substituting integer data type with int for purpose of creating hash. It is required as for now integer literals are stored as int.

# How should this be tested?
1. Create Room individual
2. Set seating capacity
3. Verify that there is no error

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/RdfLiteralHash.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/RdfLiteralHash.java
@@ -55,6 +55,10 @@ public class RdfLiteralHash {
         }else{
             if( stmt.getDatatypeURI() != null && stmt.getDatatypeURI().trim().length() > 0){
                 langOrDatatype = stmt.getDatatypeURI();
+                //Treat integer data type the same as int 
+                //With Jena 3.16.0 all integer literals are stored as int
+                //TODO: remove workaround when bug is resolved
+                langOrDatatype = replaceIntegerWithInt(langOrDatatype);
             }
         }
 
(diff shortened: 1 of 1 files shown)


bkampe (approved):
I could reproduce [issue 3728](https://github.com/vivo-project/VIVO/issues/3728). Applying this workaround led to the expected behaviour without throwing an exception.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/RdfLiteralHash.java
```diff
@@ -63,7 +67,13 @@ public static int makeRdfLiteralHash( DataPropertyStatement stmt ){
             log.debug("got hash " + hashMe.hashCode() + " for String '" + hashMe + "'");
         return hashMe.hashCode();
     }
-
+    
+    private static String replaceIntegerWithInt(String predicate) {
+        if( predicate.equals("http://www.w3.org/2001/XMLSchema#integer")) {
```
chenejac, line 72: Can you define constants for http://www.w3.org/2001/XMLSchema#integer and http://www.w3.org/2001/XMLSchema#int? It might be also useful somewhere else. Also, I am not sure we need method here, maybe ternary operator can be used instead (langOrDatatype.equals(XML_INTEGER_TYPE))?XML_INT_TYPE:langOrDatatype.

chenejac (approved):
@litvinovg it works and I approve it. However, please check one my comment whether there is any sense in what is written.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/RdfLiteralHash.java
litvinovg: I didn't create constants as I think that this code is deprecated and should be removed after we replace n3editing with dynamic API. It is a separate method for the same reasons: to emphasize that it is a temporary fix and make it easy to remove it in future. Also ternary operators harder to read and less practical for debugging.

PR #316, opened 2022-07-06
Proper fix to have the same substitutions for hashing as in RDFServiceTDB

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3728)**

Fixes issue related to creating integer literals in web interface by substituting integer data types with integer for purpose of creating hash the same way it is done in RDFServiceTDB. 
List of types to be replaced:
http://www.w3.org/2001/XMLSchema#int
http://www.w3.org/2001/XMLSchema#long
http://www.w3.org/2001/XMLSchema#short
http://www.w3.org/2001/XMLSchema#unsignedByte
http://www.w3.org/2001/XMLSchema#unsignedShort
http://www.w3.org/2001/XMLSchema#unsignedInt
http://www.w3.org/2001/XMLSchema#unsignedLong
http://www.w3.org/2001/XMLSchema#nonPositiveInteger
http://www.w3.org/2001/XMLSchema#nonNegativeInteger
http://www.w3.org/2001/XMLSchema#positiveInteger
http://www.w3.org/2001/XMLSchema#negativeInteger

# How should this be tested?
1. Create individuals
2. Add data properties with int, long, short etc.
3. Verify that there is no error

# Interested parties
@VIVO-project/vivo-committers @hudajkhan @bkampe @chenejac

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/RdfLiteralHash.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/edit/n3editing/VTwo/RdfLiteralHash.java
@@ -35,6 +35,19 @@ import edu.cornell.mannlib.vitro.webapp.dao.VitroVocabulary;
 public class RdfLiteralHash {
 
     private static final Log log = LogFactory.getLog(RdfLiteralHash.class.getName());
+    
+    private static final String integerPattern = 
+            "^http://www.w3.org/2001/XMLSchema#int$" + "|" +
+            "^http://www.w3.org/2001/XMLSchema#long$" + "|" +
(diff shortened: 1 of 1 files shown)


hudajkhan:
In general, this approach appears to be fine.  The TDB workaround has a note specifying that future versions of Jena should fix the integer conversion problem so perhaps we need to revisit in the future? But for now, this makes the web interface consistent with what is in the RDFService implementation for Jena TDB.  (For RDFService, the integer conversion seems to be used by the code that checks if graphs are equivalent.)

hudajkhan (approved):
I have not tried the conversion, but the reasoning is consistent with RDF Service implementation for TDB Jena and should not, in theory, break any of the other implementations of RDF Service.

PR #318, opened 2022-07-19
Float data properties regex

litvinovg:
Current float validation regex doesn't match values 1, 1.1, 10.1 which prevents web interface from creating data properties.

Fixes float validation regex

# How should this be tested?
* Create data property with range float
* Try adding float values: 1 , 1.1 10.1

# Interested parties
@bkampe @chenejac  @brianjlowe  @hudajkhan @gneissone @wwelling


bkampe (approved):
After applying the fix adding values like 1, 1.1, 10.1 works. It didn't work before.
grafik

hudajkhan (approved):
Reviewing the regular expressions with Georgy's help, and seeing it work based on Benjamin's comment, this seems fine.

PR #320, opened 2022-08-02
Issue/vivo 3606 : add language-specific sorting and label fields to search index

brianjlowe:
**[Issue VIVO-3606](https://github.com/vivo-project/VIVO/issues/3606)**: 

- Populates a sort field and label field in the search index for each language tag found in the list of labels for an individual.  
- If RDFService.languageFilter = true in runtime.properties, the sort field that corresponds to the current locale is used to sort the individual lists in the VClassGroup-based browse pages (People, Organizations, Events, etc.).  The original nameLowercasedSingleValue field is used as a secondary sort in case this field is not populated.
- alpha (A*) browse lookups search either for documents where the locale-specific sort field starts with the selected letter OR documents where the locale-specific sort field does not exist at all but the default nameLowerCaseSingleValued starts with the selected letter.  This should prevent content from disappearing entirely if the locale-specific sort field is not available, though it may mean that individuals appear under the wrong letter (unchanged from current behavior).
- When displaying the results of an autocomplete request, the label corresponding to the current locale is displayed if available.  If not, the original field nameRaw is used instead.

Note that this approach is intended only to enable the minimum sorting / autocomplete functionality needed by production i18nized sites.  It has (inter alia) the following key limitations:
- There is only one level of fallback (e.g. from nameLowercasedSingleValue_de-DE_s to nameLowercasedSingleValue).  There is no attempt to try de or de-AT if de-DE is not available, which means that improperly sorted values may appear if the appropriate label is not available.  Similarly, content may continue to appear under the wrong alpha heading.
- Because all sort fields use the dynamic string type _s, there is no language-specific collation enabled.  All languages will sort based on (lowercased) unicode values and not by more complex rules.  It would be nice to add this in the future, but will require the ability to modify the Solr schema according to the languages in use. 

# What's new?
- SelectQueryDocumentModifierDynamicTargetField extends SelectQueryDocumentModifier for queries that return the na e of the search index field to modify in the ?targetField variable.
- Two new document modifiers are added to home/rdf/display/everytime to populate the i18nized sort and label fields.
- AutocompleteController/IndividualListController/SearchQueryUtils are modified to take advantage of the new fields.

# How should this be tested?
- In runtime.properties, enable RDFService.languageFilter = true and add en_US, es, fr_CA, and de_DE as the selectable locales.
- Load sorttest.n3.txt (attached).
- Switch between the four locales and observe that the items in the People tab sort and alpha-filter properly, displaying the language-appropriate label in parentheses except in the case of 'Yanny' when de_DE is selected.  In the latter case, Yanny (en-US) will be displayed instead.
- Observe that Yanny is still browsable on the People tab when the locale is set to de_DE, even though there is no de-DE label for the individual.
- Add a publication to the DB.  Add an author.  Autocomplete on the author names 'Alpha', 'Bravo', 'Charlie' or 'Delta'.  Note that all 4 individuals are returned when you type one of these names.  This is because the autocomplete edgengram is not language-specific, and is out of scope for this PR.  The improvement with this PR is that the labels in the autocomplete dropdown will change according to your currently-selected locale.

# Interested parties
@VIVO-project/vivo-committers

[sorttest.n3.txt](https://github.com/vivo-project/Vitro/files/9241040/sorttest.n3.txt)

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/IndividualListController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/IndividualListController.java
@@ -6,9 +6,11 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
-import edu.cornell.mannlib.vitro.webapp.web.templatemodels.individuallist.ListedIndividualBuilder;
+import javax.servlet.annotation.WebServlet;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -16,6 +18,7 @@ import org.apache.commons.logging.LogFactory;
 import edu.cornell.mannlib.vitro.webapp.beans.Individual;
 import edu.cornell.mannlib.vitro.webapp.beans.VClass;
 import edu.cornell.mannlib.vitro.webapp.beans.VClassGroup;
+import edu.cornell.mannlib.vitro.webapp.config.ConfigurationProperties;
 import edu.cornell.mannlib.vitro.webapp.controller.VitroRequest;
 import edu.cornell.mannlib.vitro.webapp.controller.freemarker.responsevalues.ExceptionResponseValues;
 import edu.cornell.mannlib.vitro.webapp.controller.freemarker.responsevalues.ResponseValues;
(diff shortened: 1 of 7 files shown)


brianjlowe:
Based on offline discussion with @litvinovg it seems that the limitations of the approach, particularly the need for labels to be consistently populated for all languages across all browseable individuals, may make this impractical for many users.

My initial goal was to do this without requiring Solr schema reload, but it may be better to have a companion vivo-solr PR that configures the sort and label fields for each language with the appropriate collation/filters.  The document modifier would then take each of the configured selectable locales and populate its fields consistently on every record.

brianjlowe:
Superseded by https://github.com/vivo-project/Vitro/pull/321

PR #321, opened 2022-08-05
Issue/vivo 3606 : add language-specific sorting and label fields to search index

litvinovg:
**[Issue VIVO-3606](https://github.com/vivo-project/VIVO/issues/3606)**: 
This PR is a successor of this [PR](https://github.com/vivo-project/Vitro/pull/320)

[VIVO-Solr PR](https://github.com/vivo-project/vivo-solr/pull/6) (companion PR, but it still works without it )

- Populates a sort field and label field in the search index for each language tag found in the list of labels for an individual.  
- If RDFService.languageFilter = true in runtime.properties, the sort field that corresponds to the current locale is used to sort the individual lists in the VClassGroup-based browse pages (People, Organizations, Events, etc.).  The original nameLowercasedSingleValue field is used as a secondary sort in case this field is not populated.
- alpha (A*) browse lookups search either for documents where the locale-specific sort field starts with the selected letter OR documents where the locale-specific sort field does not exist at all but the default nameLowerCaseSingleValued starts with the selected letter.  This should prevent content from disappearing entirely if the locale-specific sort field is not available, though it may mean that individuals appear under the wrong letter (unchanged from current behavior).
- When displaying the results of an autocomplete request, the label corresponding to the current locale is displayed if available.  If not, the original field nameRaw is used instead.

Note that this approach is intended only to enable the minimum sorting / autocomplete functionality needed by production i18nized sites.  It has the following key limitations:
- There is only one level of fallback (e.g. from de-DE_label_sort to nameLowercasedSingleValue).  There is no attempt to try de or de-AT if de-DE is not available, which means that improperly sorted values may appear if the appropriate label is not available.  Similarly, content may continue to appear under the wrong alpha heading.
- Because all sort fields use the dynamic string type, there is no language-specific collation enabled.  All languages will sort based on (lowercased by solr schema configuration) unicode values and not by more complex rules.  It would be nice to add this in the future, but will require the ability to modify the Solr schema according to the languages in use. 

# What's new?
- SelectQueryDocumentModifierDynamicTargetField extends SelectQueryDocumentModifier for queries that performs queries for each of locales found in runtime properties and return search index fields with names composed of locale + targetSuffix.
- Two new document modifiers are added to home/rdf/display/everytime to populate the i18nized sort and label fields.
- AutocompleteController/IndividualListController/SearchQueryUtils are modified to take advantage of the new fields.
- If solr schema doesn't contain fields with "_label_sort" and "_label_display" suffixes then dynamic fields with standard string type will be created at SolrSearchEngine startup. It will fix label display and label sort on instances where old solr schema is used (still adviced to be updated ).

# How should this be tested?
- In runtime.properties, enable RDFService.languageFilter = true and add en_US, es, fr_CA, and de_DE as the selectable locales.
- Load sorttest.n3.txt and sorttest_2.n3.txt (attached).
- Switch between the four locales and observe that the items in the People tab sort and alpha-filter properly, displaying the language-appropriate label in parentheses except in the case of 'Yanny' when de_DE is selected.  In the latter case, Yanny (en-US) will be displayed instead.
- Observe that Yanny is still browsable on the People tab when the locale is set to de_DE, even though there is no de-DE label for the individual.
- Add a publication to the DB.  Add an author.  Autocomplete on the author names 'Alpha', 'Bravo', 'Charlie' or 'Delta'.  Note that all 4 individuals are returned when you type one of these names.  This is because the autocomplete is not language-specific, and is out of scope for this PR.  The improvement with this PR is that the labels in the autocomplete dropdown will change according to your currently-selected locale.
- Verify that individuals without labels for specified languages still sorted correctly.
- Try load VIVO with old and new solr schemas and run rebuild search index. With old solr schema default dynamic fields should be created. With updated solr schema dynamic fields provided by solr schema should be used.

# Interested parties
@VIVO-project/vivo-committers

[sorttest.n3.txt](https://github.com/vivo-project/Vitro/files/9241040/sorttest.n3.txt)
[sorttest_2.n3.txt](https://github.com/vivo-project/Vitro/files/9267085/sorttest_2.n3.txt)

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/IndividualListController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/IndividualListController.java
@@ -6,9 +6,11 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
-import edu.cornell.mannlib.vitro.webapp.web.templatemodels.individuallist.ListedIndividualBuilder;
+import javax.servlet.annotation.WebServlet;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -16,6 +18,7 @@ import org.apache.commons.logging.LogFactory;
 import edu.cornell.mannlib.vitro.webapp.beans.Individual;
 import edu.cornell.mannlib.vitro.webapp.beans.VClass;
 import edu.cornell.mannlib.vitro.webapp.beans.VClassGroup;
+import edu.cornell.mannlib.vitro.webapp.config.ConfigurationProperties;
 import edu.cornell.mannlib.vitro.webapp.controller.VitroRequest;
 import edu.cornell.mannlib.vitro.webapp.controller.freemarker.responsevalues.ExceptionResponseValues;
 import edu.cornell.mannlib.vitro.webapp.controller.freemarker.responsevalues.ResponseValues;
@@ -27,8 +30,7 @@ import edu.cornell.mannlib.vitro.webapp.modules.searchEngine.SearchEngineExcepti
 import edu.cornell.mannlib.vitro.webapp.modules.searchEngine.SearchQuery;
 import edu.cornell.mannlib.vitro.webapp.utils.searchengine.SearchQueryUtils;
 import edu.cornell.mannlib.vitro.webapp.web.templatemodels.individuallist.ListedIndividual;
-
-import javax.servlet.annotation.WebServlet;
+import edu.cornell.mannlib.vitro.webapp.web.templatemodels.individuallist.ListedIndividualBuilder;
 
 /**
  * Generates a list of individuals for display in a template
@@ -43,6 +45,7 @@ public class IndividualListController extends FreemarkerHttpServlet {
     private static final int MAX_PAGES = 40;  // must be even
 
     private static final String TEMPLATE_DEFAULT = "individualList.ftl";
+    private static final String LANGUAGE_FILTER_PROPERTY = "RDFService.languageFilter";
 
     @Override
     protected ResponseValues processRequest(VitroRequest vreq) {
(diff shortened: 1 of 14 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchindex/documentBuilding/SelectQueryDocumentModifierDynamicTargetField.java
```diff
@@ -0,0 +1,110 @@
...
+ * Target field names are composed of locale + fieldSuffix.
+ * 
+ * Each query should contain a ?uri variable, which will be replaced by the URI
+ * of the individual.
+ * 
+ * All of the other result fields in each row of each query will be converted to
+ * strings and added to the field.
+ *
+ */
+public class SelectQueryDocumentModifierDynamicTargetField extends SelectQueryDocumentModifier
```
wwelling, line 39: Is this named properly?

litvinovg: Renamed

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/IndividualListController.java
```diff
@@ -152,11 +155,17 @@ public static int getPageParameter(VitroRequest request) {
...
+            int page, String alpha, VitroRequest vreq)
     throws SearchException{
    	 	try{
+            ConfigurationProperties props = ConfigurationProperties.getBean(vreq);
+            boolean languageFilter = Boolean.valueOf(props.getProperty(
+                    LANGUAGE_FILTER_PROPERTY, "false"));
             List<String> classUris = Collections.singletonList(vclassURI);
-			IndividualListQueryResults results = buildAndExecuteVClassQuery(classUris, alpha, page, INDIVIDUALS_PER_PAGE, vreq.getWebappDaoFactory().getIndividualDao());
+			IndividualListQueryResults results = buildAndExecuteVClassQuery(
+			        classUris, alpha, ((languageFilter) ? vreq.getLocale() : null),
```
chenejac, line 167: Can we move this code inside buildAndExecuteVClassQuery 
((languageFilter) ? vreq.getLocale() : null)
I mean you are repeating this code at two invocation, maybe you can beside method
buildAndExecuteVClassQuery(
			List<String> vclassURIs, String alpha, Locale locale, int page,
			int pageSize, IndividualDao indDao) {...}
also create
buildAndExecuteVClassQuery(
			List<String> vclassURIs, String alpha, boolean languageFilter, int page,
			int pageSize, IndividualDao indDao){
return buildAndExecuteVClassQuery(vclassURIs, alpha, ((languageFilter) ? vreq.getLocale() : null), page, pageSize, indDao);
}

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/base/BaseSearchQuery.java
```diff
@@ -7,6 +7,7 @@
 import java.util.Collections;
 import java.util.HashMap;
```
chenejac, line 8: Is this import needed anymore (HashMap)

chenejac (approved):
@litvinovg can you please read my comments before we merge your PR. Great contribution!!!

api/src/main/java/edu/cornell/mannlib/vitro/webapp/search/VitroSearchTermNames.java
```diff
@@ -70,5 +70,11 @@ public class VitroSearchTermNames {
 
 	/** Source institution name */
 	public static final String SITE_NAME = "siteName";
+	
+	/** Multilingual sort field suffix */
+	public static final String SORT_SUFFIX = "_label_sort";
+	
+	/** Multilingual label field suffix */
+	public static final String LABEL_DISPLAY_SUFFIX = "_label_display";
```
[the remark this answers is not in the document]
litvinovg: done

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/IndividualListController.java
litvinovg: done

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/base/BaseSearchQuery.java
litvinovg: removed unused import

chenejac (approved):
Thanks for the contribution.

matthiasluehr (approved):
I tested these changes and they worked in my development environment. Individuals appeared on the expected index pages dependent on the selected language. Autocompleter was also aware of different languages and multiple labels. Tested with provided sample data as well as some own examples.

wwelling (approved):
Looks good to me.

PR #322, opened 2022-08-09
fix: use public description in data property template model

litvinovg:
[Issue](https://github.com/vivo-project/VIVO/issues/3736)
Uses public description for data properties when verbose property display is on
Before the fix public description show up only for object properties and faux properties

# How should this be tested?

1. Log in
2. Turn on verbose property display
3. Hover over data property  ( over **data** word ) that has public description (A hint with public description should appear)

# Interested parties
@VIVO-project/vivo-committers


chenejac (approved):
Works for me

brianjlowe (approved):

PR #334, opened 2022-10-11
[ i18n sprint ] Decouple email templates and translations

litvinovg:
[Companion Vitro-languages PR](https://github.com/vivo-project/Vitro-languages/pull/61)
Modified FreemarkerEmailMessage and classes that used it:
Removed EmailDirective and error-email.ftl
Send email in separate thread to prevent web interface delays.
With this PR translations, extracted from email templates into property files could be used.
#How to test#
1. Configure email settings
For each of installed languages execute following actions:
3. Create new user, e-mail to the user about created account should be received.
4. Use the link from the e-mail, set user password, you should receive e-mail after password is set.
5. Try reset user's password, e-mail to the user about created account should be received.
6. Try to log in as a user and change e-mail address, you should receive e-mail about that.
7. Break your page-home.ftl and open home page, you should see "There was an error in the system. This error has been reported to the site administrator. " Check email in administrator's mailbox

# Interested parties
 @VIVO-project/vivo-committers


chenejac (approved):

chenejac (approved):

api/src/main/java/edu/cornell/mannlib/vitro/webapp/email/FreemarkerEmailMessage.java
```diff
@@ -182,15 +208,27 @@ public boolean send() {
...
+			sendMessage(msg);
 			return true;
 		} catch (MessagingException e) {
 			log.error("Failed to send message.", e);
 			return false;
 		}
 	}
 
+	private void sendMessage(MimeMessage msg) {
+		Thread thread = new Thread() {
```
brianjlowe, line 220: Why the asynchronous send?  I think we already have a problem where the GUI claims that emails were sent successfully even if errors occurred.  It seems that this might make that harder to address.

brianjlowe (commented):
This looks good to me overall.  I don't have an easy way of testing email transmission with real SMTP at the moment, but am inclined to approve the PR.  One question inline regarding the asynchronous transmission.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/email/FreemarkerEmailMessage.java
litvinovg: Before that if smtp server have some delays 20-120 sec (it is one of greylisting approaches) then web interface will freeze for that period. 
I agree, current notification about successfully sent email is misleading. I guess It should be replaced with something like "The message is going to be sent shortly" and an option to look up the results in notifications journal.

brianjlowe: Sounds reasonable.  I guess I had never encountered delays that long, but that would certainly be something to avoid.  We should have a better way of verifying that emails were sent, but that seems out of scope for this PR.

brianjlowe (approved):

brianjlowe:
Not merging yet pending approval of companion PR.

PR #337, opened 2022-10-18
[Dynapi] Resource generator

litvinovg:
# What's new?
Created resource generator with configurable uri scheme. Could be used as a deserializer method in parameter definition.

# How should this be tested?
Unit test included

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
@@ -1,12 +1,7 @@
 package edu.cornell.mannlib.vitro.webapp.dynapi.data.implementation;
 
 import java.math.BigInteger;
-import java.util.Random;
 import java.util.UUID;
-
-
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import org.apache.jena.rdf.model.Resource;
 import org.apache.jena.rdf.model.impl.ResourceImpl;
 import edu.cornell.mannlib.vitro.webapp.dynapi.data.conversion.InitializationException;
@@ -15,11 +10,15 @@ import edu.cornell.mannlib.vitro.webapp.dao.WebappDaoFactory;
 
 public class ResourceGenerator {
 
-	private static final Log log = LogFactory.getLog(ResourceGenerator.class.getName());
-
-	protected static Object JAVA_UUID = "java_uuid";
-	protected static Object OLD_NUMBER = "old_number";
-	protected static Object UUID_NUMBER = "uuid_number";
+	private static final String BASE62_CHARS = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+	protected static String JAVA_UUID = "java_uuid";
+	protected static String JAVA_UUID_NO_DASH = "java_uuid_no_dash";
+	protected static String UUID_NUMBER = "uuid_number";
+	protected static String UUID_BASE62 = "uuid_base62";
+	
+	private final String LEAD_ZEROS = "00000000000000000000000";
+	private BigInteger base = BigInteger.valueOf(62);
+	private BigInteger zero = BigInteger.valueOf(0);
 
 	private static ResourceGenerator INSTANCE = new ResourceGenerator();
 
@@ -34,17 +33,15 @@ public class ResourceGenerator {
 	}
 
 	/**
-	 * Creates a resource from input comma separated list of uri part names Values
-	 * of each parts calculated independently and concatenated result returned If
-	 * part name is not defined in the class, then it is added to the uri name If
-	 * part name is {@value #JAVA_UUID}, then new generated UUID will be used as a
-	 * part If part name is {@value #UUID_NUMBER}, then new generated UUID
-	 * represented as a string of 40 numbers will be used as a part If part name is
-	 * {@value #OLD_NUMBER} and it is the last one, then old individual uri should
-	 * be created. All previously defined parts will be used as a new namespace.
+	 * Creates a resource from input comma separated list of keys.
+	 * Values of each parts calculated independently and concatenated result returned.
+	 * If part name is not known, then it will be returned as a value.
+	 * Key {@value #JAVA_UUID} returns value in format [0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}
+	 * Key {@value #JAVA_UUID_NO_DASH} returns value in format [0-9a-f]{32}
+	 * Key {@value #UUID_NUMBER} returns value in format [0-9]{40} 
+ 	 * Key {@value #UUID_BASE62} returns value in format [0-9a-zA-Z]{23} 
 	 * @throws InitializationException 
 	 */
-
 	public Resource getUriFromFormat(String input) throws InitializationException {
 		String uri = getUri(input);
 		Resource res = new ResourceImpl(uri);
@@ -56,9 +53,6 @@ public class ResourceGenerator {
 		StringBuilder uriBuilder = new StringBuilder();
 		for (int i = 0; i < parts.length; i++) {
 			String partName = parts[i].trim();
-			if (isOldFormatRequested(parts, i, partName) && wadf != null) {
-				return getOldURI(uriBuilder.toString());
-			}
 			uriBuilder.append(getUriPart(partName));
 		}
 		if (uriBuilder.length() == 0) {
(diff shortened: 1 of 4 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	private String getUri(String input) throws InitializationException {
+		String[] parts = input.split("[\\s,]+");
+		StringBuilder uriBuilder = new StringBuilder();
+		for (int i = 0; i < parts.length; i++) {
+			String partName = parts[i].trim();
+			if (isOldFormatRequested(parts, i, partName) && wadf != null) {
+				return getOldURI(uriBuilder.toString());
+			}
```
chenejac, line 61:
```suggestion

```

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+			uriBuilder.append(getUriPart(partName));
```
chenejac, line 62:
```suggestion
			uriBuilder.append(getUriPart(uriBuilder.toString(), partName));
```

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+		String uri = uriBuilder.toString();
+		if (wadf != null && wadf.checkURI(uri) != null) {
+			throw new InitializationException("Uri check failed");
+		}
+		return uri;
+	}
+
+	private boolean isOldFormatRequested(String[] parts, int i, String part) {
+		return parts.length == i + 1 && OLD_NUMBER.equals(part);
+	}
```
chenejac, line 76:
```suggestion

```

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+
+	private static String getUriPart(String keyword) {
```
chenejac, line 78:
```suggestion
	private static String getUriPart(String namespace, String keyword) {
```

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+	private static String getJavaUUID() {
+		return UUID.randomUUID().toString();
+	}
+
+	public void init(WebappDaoFactory webappDaoFactory) {
+		wadf = webappDaoFactory;
+	}
+
+	private String getOldURI(String newNamespace) throws InitializationException {
+		String uri = null;
```
chenejac, line 107:
```suggestion
		if (wadf == null)
		     return getJavaUUIDNumber();
		String uri = null;
		String uriNumber = null;
```

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+		String errMsg = null;
+		Random random = new Random();
+		boolean uriIsGood = false;
+		int attempts = 0;
+
+		while (!uriIsGood && attempts < 30) {
+			uri = newNamespace + random.nextInt(Math.min(Integer.MAX_VALUE, (int) Math.pow(2, attempts + 13)));
```
chenejac, line 114:
```suggestion
			uriNumber = random.nextInt(Math.min(Integer.MAX_VALUE, (int) Math.pow(2, attempts + 13)));
			uri = newNamespace + uriNumber;
```

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+			if (errMsg != null) {
+				uri = null;
+			} else {
+				uriIsGood = true;
+			}
+			attempts++;
+		}
+		if (uri == null) {
+			log.error("Generated uri is null, fallback to generate uri in new format");
+			uri = newNamespace + getJavaUUIDNumber();
```
chenejac, line 125:
```suggestion
			uriNumber = getJavaUUIDNumber();
```

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+		}
+		if (wadf.checkURI(uri) != null) {
+			throw new InitializationException("Uri check failed");
+		}
+
+		return uri;
```
chenejac, line 131:
```suggestion
		return uriNumber;
```

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+	private static String getUriPart(String keyword) {
+		if (JAVA_UUID.equals(keyword)) {
+			return getJavaUUID();
+		}
+		if (UUID_NUMBER.equals(keyword)) {
+			return getJavaUUIDNumber();
+		}
+		//If wadf is not available fallback to use of UUID_NUMBER
+		if (OLD_NUMBER.equals(keyword)) {
+			return getJavaUUIDNumber();
```
chenejac, line 87:
```suggestion
			return getOldURI(namespace);
```

chenejac (changes requested):
@litvinovg this works for me. However, I don't like special case for the old uri approach. Can you please check my suggestion for different organization of the code? What do you think about that?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,134 @@
...
+		if (uri == null) {
+			log.error("Generated uri is null, fallback to generate uri in new format");
+			uri = newNamespace + getJavaUUIDNumber();
+		}
+		if (wadf.checkURI(uri) != null) {
+			throw new InitializationException("Uri check failed");
+		}
+
+		return uri;
+	}
```
chenejac, line 132:
```suggestion
	private String getUri(String input) throws InitializationException {
		String[] parts = input.split("[\\s,]+");
		StringBuilder uriBuilder = new StringBuilder();
		for (int i = 0; i < parts.length; i++) {
			String partName = parts[i].trim();
			uriBuilder.append(getUriPart(uriBuilder.toString(), partName));
		}
		if (uriBuilder.length() == 0) {
			uriBuilder.append(getJavaUUID());
		}
		String uri = uriBuilder.toString();
		if (wadf != null && wadf.checkURI(uri) != null) {
			throw new InitializationException("Uri check failed");
		}
		return uri;
	}

	private static String getUriPart(String namespace, String keyword) {
		if (JAVA_UUID.equals(keyword)) {
			return getJavaUUID();
		}
		if (UUID_NUMBER.equals(keyword)) {
			return getJavaUUIDNumber();
		}
		if (OLD_NUMBER.equals(keyword)) {
			return getOldURI(namespace);
		}
		return keyword;
	}

	private static String getJavaUUIDNumber() {
		final UUID randomUUID = UUID.randomUUID();
		final String base16String = randomUUID.toString().replace("-", "");
		return String.format("%040d", new BigInteger(base16String, 16));
	}

	private static String getJavaUUID() {
		return UUID.randomUUID().toString();
	}

	public void init(WebappDaoFactory webappDaoFactory) {
		wadf = webappDaoFactory;
	}

	private String getOldURI(String newNamespace) throws InitializationException {
		if (wadf == null)
		    return getJavaUUIDNumber();
		String uri = null;
		String uriNumber = null;
		String errMsg = null;
		Random random = new Random();
		boolean uriIsGood = false;
		int attempts = 0;

		while (!uriIsGood && attempts < 30) {
			uriNumber = random.nextInt(Math.min(Integer.MAX_VALUE, (int) Math.pow(2, attempts + 13)));
			uri = newNamespace + uriNumber;
			errMsg = wadf.checkURI(uri);
			if (errMsg != null) {
				uriNumber = null;
				uri = null;
			} else {
				uriIsGood = true;
			}
			attempts++;
		}
		if (uriNumber == null) {
			log.error("Generated uri is null, fallback to generate uri in new format");
			uriNumber = getJavaUUIDNumber();
		}
		/*
		I think this is not needed, because it is checked at the end of the getUri method
		if (wadf.checkURI(uri) != null) {
			throw new InitializationException("Uri check failed");
		}*/

		return uriNumber;
	}
```

chenejac (changes requested):
@litvinovg I have tried to create one big code suggestion instead of lot of small suggestions. Please, check if it is improving the readability of the code. Anyway, I am not sure should we support three different uri generation, or we should select one and force that to be used in the next minor version of VIVO.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
litvinovg: This solution still force us to mix new approach of constructing URIs from user defined schemes and old approach.
I think I made a mistake trying to mix them. 
So I suggest to drop support of old approach, as it is not only rigid, but also more expensive and produces less unique uri's.
We should decide what should be our standard, but I think it makes sense to leave this easy configurable.

chenejac (changes requested):
@litvinovg please check my comment

api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/data/implementation/ResourceGenerator.java
```diff
@@ -0,0 +1,151 @@
...
+		String uri = null;
+		String errMsg = null;
+		Random random = new Random();
+		boolean uriIsGood = false;
+		int attempts = 0;
+		int number = 0;
+		while (!uriIsGood && attempts < 30) {
+			number = random.nextInt(Math.min(Integer.MAX_VALUE, (int) Math.pow(2, attempts + 13)));
+			uri = prefix + number;
+			errMsg = wadf.checkURI(uri);
```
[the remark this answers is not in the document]
litvinovg: Thanks! Fixed

chenejac (approved):

chenejac:
superseded by https://github.com/vivo-project/Vitro/pull/364

PR #342, opened 2022-10-30
Replacement of language comparisons using '=' by the 'langMatches'

michel-heon:
see also: [PR3789](https://github.com/vivo-project/VIVO/pull/3789)


wwelling (approved):
Not the cleanest of files, but the change looks good. https://www.w3.org/TR/sparql11-query/#func-langMatches

PR #351, opened 2022-11-29
[3800 i18n redesign] Supporting private tags

chenejac:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues)**: [3800](https://github.com/vivo-project/VIVO/issues/3800)

[Companion PR](https://github.com/vivo-project/VIVO/pull/3802)

Adding support for local customization of UI labels by using language tags. It means an institution can add in the directory <VIVO HOME>/rdf/i18n/ its own local customization of translations. Those local, institution-specific translations should be in the directory which name is aligned with language tag specification for private tags (for instance fr_CA_x_uqam or de_DE_x_tib). The idea is similar as for property files and usage of <tomcat>webapps/i18n/local directory. The institution can easily update VIVO version, and continue to use previously customized labels and vocabularies. 

# What's new?
- Adding support for loading of language tag with private subtags (x) in the LocaleUtility.java
- Replacing locale.toString() with locale.toLanguageTag():
`locale.toLanguageTag()` 
produces
`fr-CA-x-uqam`
while the code
`locale.toString()`
produces
`fr_CA_#x-uqam`
We have code which do replacement of _ and - in string representation of locale, therefore, the first one string is needed
- Introduce checking for locale.stripExtensions() when it is needed:
For locale `fr-CA-x-uqam` it produces `fr-CA`, the first one is extension of the second, and usually it includes all labels for the language `fr-CA` plus some extensions for specific cases. 

# How should this be tested?
1. Install VIVO
2. Unzip [this file](https://github.com/vivo-project/Vitro/files/10185798/fr_CA_x_uqam.zip) into <VIVO HOME>/rdf/i18n/
3. Add into runtime properties 
`languages.selectableLocales = en_US, fr_CA, fr_CA_x_uqam`
4. Run tomcat and open VIVO in a web browser
5. Change language between the second and third language in the list
6. At the home page you should found the label "Recherche" and "Recherche uqam" depends on selection of the French variant.
7. If you login, you will also notice that logout option in the menu has been changed to "Déconnexion wilma uqam" when the last language has been selected
8. Also, on some pages labels with uqam sufix will appear. 

Example:
* Does this change require documentation to be updated? **Yes, it should be documented how this feature should be used.**
* Does this change add any new dependencies? **No**
* Does this change require any other modifications to be made to the repository? **No**
* Could this change impact execution of existing code? **I don't think so**
* Large pull requests should be avoided. If this PR is large (more than 1,000 lines of codes), please provide short explanation why your contribution can't be decoupled in smaller PRs. **It is a small PR.** 

# Interested parties

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/ViewLabelsServlet.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/ViewLabelsServlet.java
@@ -170,7 +170,7 @@ public class ViewLabelsServlet extends FreemarkerHttpServlet{
 			Locale currentLocale) throws FileNotFoundException {
 		HashMap<String, String> map = new HashMap<String, String>();
 		//Replacing the underscore with a hyphen because that is what is represented in the actual literals
-		map.put("code", locale.toString().replace("_", "-"));
+		map.put("code", locale.toLanguageTag().replace("_", "-"));
 		map.put("label", locale.getDisplayName(currentLocale));
 		return map;
 	}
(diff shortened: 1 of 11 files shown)


michel-heon:
### Test result:
On my side. this PR works according to the requested checks. 
Suggestion: it would be nice, in the menu of the choice of the liguistic context to indicate also the sub context 'uqam' in the label in order to better distinguish it 
ex.: 
menu

> Select a language
> - English (United State)
> - French (Canada)
> - French (Canada) uqam

michel-heon (approved):

litvinovg (approved):
I would suggest to always use curly braces for all if statements.
It works fine.

chenejac:
> I would suggest to always use curly braces for all if statements. It works fine.

braces added to all if statements

PR #356, opened 2022-12-16
support for combining parameters and apostrophe

chenejac:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues)**: [3393](https://github.com/vivo-project/VIVO/issues/3393)

Please, find more about the issue also at this [discussion](https://stackoverflow.com/questions/17569608/format-a-message-using-messageformat-format-in-java)

Add support for combining apostrophe and parameters:
`Taille maximale de l'image: {0} mégaoctets`

It might be resolved in i18n label files by replacing ' with '', and in some cases it has been [done](https://github.com/vivo-project/Vitro/blob/i18n-redesign/home/src/main/resources/rdf/i18n/en_US/interface-i18n/firsttime/vitro_UiLabel.ttl#L2642), but not in [all cases](https://github.com/vivo-project/Vitro/blob/i18n-redesign/home/src/main/resources/rdf/i18n/en_US/interface-i18n/firsttime/vitro_UiLabel.ttl#L5178). However, this might be also resolved in java files when we are invoking MessageFormat.format method. This PR should support both cases
`Taille maximale de l'image: {0} mégaoctets`
and
`Taille maximale de l''image: {0} mégaoctets`

# What's new?
In the invocation of MessageFormat.format method, firstly all double apostrophe ('') are replaced with single, then single apostrophe is replaced with double. At the end, '' and ' are ''

# How should this be tested?

- Run VIVO application
- Log in
- Open a researcher profile page
- Click on uploading the image
- Change language to French
- Check labels `Taille maximale de l'image: 6 mégaoctets`, `Taille minimale de l'image: 200 par 200 pixels`

# Interested parties

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/TranslationProvider.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/TranslationProvider.java
@@ -165,7 +165,7 @@ public class TranslationProvider {
 		if (parameters.length == 0) {
 			return textString;
 		} else {
-			return MessageFormat.format(textString, parameters);
+			return MessageFormat.format(textString.replaceAll("''", "'").replaceAll("'", "''"), parameters);
 		}
 	}
 
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/TranslationProvider.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/TranslationProvider.java
@@ -165,7 +165,21 @@ public class TranslationProvider {
 		if (parameters.length == 0) {
 			return textString;
 		} else {
-			return MessageFormat.format(textString.replaceAll("''", "'").replaceAll("'", "''"), parameters);
+			return MessageFormat.format(TranslationProvider.preprocessForFormating(textString), parameters);
+		}
+	}
+
+	/**
+	 * This method should prepare the inputText for MessageFormat.format method. At the moment it is replacing single
+	 * apostrophe with double, it might be extented in the future with some additional preprocessing.
+	 * @param inputText - string which should be preprocessed
+	 * @return preprocessed input string, i.e. string with replaced single apostrophe with double
+	 */
+	public static String preprocessForFormating(String inputText){
+		if (inputText != null) {
+			return inputText.replace("''", "'").replace("'", "''");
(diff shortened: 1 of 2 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/TranslationProvider.java
```diff
@@ -165,7 +165,7 @@ private static String formatString(String textString, Object... parameters) {
 		if (parameters.length == 0) {
```
chenejac, line 165: While I was debugging I noticed that in some cases when {0} are used in the label translation, parameters.length is zero here, but later in I18nStringTemplateModel it is replaced as it should be (for instance for Taille minimale de l'image: {0} par {1} pixels). I hope @litvinovg is aware of this, and this is working in accordance with his expectations.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/freemarker/I18nStringTemplateModel.java
```diff
@@ -83,13 +83,13 @@ public Object exec(List args) throws TemplateModelException {
...
 	 * @param preProcessed String "startSep + key + intSep + textString + intSep + message + endSep"
-	 * @param arguments that should be listed before message and substituted in the message itself 
+	 * @param args that should be listed before message and substituted in the message itself
 	 * @return
 	 */
 	private String getOnlineTranslationsFormattedMessage(String preProcessed, Object[] args) {
 		String[] parts = preProcessed.split(I18nBundle.INT_SEP);
 		final int messageIndex = parts.length -1;
-		String message = MessageFormat.format(parts[messageIndex], args);
+		String message = MessageFormat.format(parts[messageIndex].replaceAll("''", "'").replaceAll("'", "''"), args);
```
chenejac, line 92: @litvinovg please check do we need this here

litvinovg: I think replace should be used instead of replaceAll as there is no regex used.
I suggest to extract method that does it instead of inline fixes.

litvinovg (approved):
Fixes the issue.
Looks good.

PR #362, opened 2023-02-01
Fix for Russian translation of email temlate

litvinovg:
Fix for Russian translation of email temlate
Added link to VIVO instance.

# Interested parties
@wwelling @chenejac


chenejac (approved):
I am not a Russian speaker, but I can confirm it is extended (in plain and html) with a password link. Not linked only with Russian, but I am wondering why in html_text we are combining html tags and \n, instead of using `<br/>`? Anyway, it is working as expected.

wwelling (approved):
I cannot read the text but looks to include the link as expected.

tawahle (approved):

PR #363, opened 2023-02-07
fix creation of faux property with xml literal range value

litvinovg:
Added some safety checks in FauxPropertyRetryController
# How should this be tested?
Create faux property for base property with XMLLiteral range
# Interested parties
@wwelling @chenejac


chenejac (approved):
I have tested this and there is no more exception. Code looks good.

wwelling (approved):
It works!

image

PR #365, opened 2023-02-09
[Issue 3767] Replacement of language comparisons using '=' by the 'langMatches' (without whitespace changes)

wwelling:
see also: [PR3789](https://github.com/vivo-project/VIVO/pull/3789)


chenejac (approved):

api/src/main/java/edu/cornell/mannlib/vitro/webapp/services/shortview/FakeApplicationOntologyService.java
```diff
@@ -479,7 +482,7 @@ private static OntModel initializeFakeDisplayModel() {
 			Property saveToVarProperty = m
 					.getProperty(DisplayVocabulary.SAVE_TO_VAR);
 
-			m.add(dataGetter, queryProperty, QUERY_STRING_LANG); //UQAM-Optimization Using query with linguistic context
+			m.add(dataGetter, queryProperty, QUERY_STRING_LANG("en-US")); //UQAM-Optimization Using query with linguistic context
```
wwelling, line 485: I am wondering, is this correct to use a hard coded `en-US`. The calling method `initializeFakeDisplayModel` and the method `QUERY_STRING_LANG` are both static and cannot reference the non-static class member `langCtx`. Previously it appeared to template the SPARQL query using a data context map with `langCtx`.

I have approved this previously but on second thought, am not sure it is correct. Yet the class name does not indicate anything of importance and the comments appear to indicate specfic use casees.

chenejac (approved):
@michel-heon confirmed this is working for AWS Neptun

PR #368, opened 2023-02-22
Infrastructure improvements - Migration to Java 11 from Java 8

michel-heon:
See also https://github.com/vivo-project/VIVO/pull/3836 for documentation
@chenejac

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -45,6 +45,7 @@
                                 <echo message="[java.vendor.version]        ${java.vendor.version}" />
                                 <echo message="[java.version]               ${java.version}" />
                                 <echo message="[java.specification.version] ${java.specification.version}" />
+                                <echo message="[maven.compiler.release]     ${maven.compiler.release}" />
                                 <echo message="###################################################################" />
                                 <echo message="###################################################################" />
                             </target>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -24,30 +24,34 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-antrun-plugin</artifactId>
-                <version>1.1</version>
+                <version>3.1.0</version>
                 <executions>
                     <execution>
+                        <id>echo-properties</id>
                         <phase>validate</phase>
                         <goals>
                             <goal>run</goal>
                         </goals>
                         <configuration>
-                            <tasks>
-                                <echo>###################################################################</echo>
-                                <echo>Displaying value of 'VIVO build' property</echo>
-                                <echo>###################################################################</echo>
-                                <echo>[vitro-version] ${vitro-version}</echo>
-                                <echo>[vivo-version] ${vivo-version}</echo>
-                                <echo>[vivo-dir] ${vivo-dir}</echo>
-                                <echo>[project.version] ${project.version}</echo>
-                                <echo>###################################################################</echo>
-                                <echo>###################################################################</echo>
-                            </tasks>
+                            <skip>false</skip>
+                            <target>
+                                <echo message="###################################################################" />
+                                <echo message="Displaying value of '${project.artifactId}' properties" />
+                                <echo message="###################################################################" />
+                                <echo message="[vitro-version]              ${vitro-version}" />
+                                <echo message="[vivo-version]               ${vivo-version}" />
+                                <echo message="[vivo-dir]                   ${vivo-dir}" />
+                                <echo message="[project.version]            ${project.version}" />
+                                <echo message="[java.vendor.version]        ${java.vendor.version}" />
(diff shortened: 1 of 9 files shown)


litvinovg (changes requested):
@michel-heon please remove VIVO an VItro version change from pom files. It is not related to the PR scope and can cause merge conflicts in future.

michel-heon:
@litvinovg The return to 1.13.1 is done.... but this version corresponds to the previous sprint and is not representative of the evolution of development. This version misalignment may lead to misunderstandings about the evolution of the development.

litvinovg (changes requested):
I decided to try it out on some OS where Java 8 was installed by default. I tried Centos 7.
But the response I received after building it with Java 8 was confusing. It wasn't about Java requirement, it was about new dependency on maven-antrun-plugin that is not provided with default (3.0.5) maven version on Centos 7.
[INFO] ------------------------------------------------------------------------ [ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (echo-properties) on project vitro-api: The plugin org.apache.maven.plugins:maven-antrun-plugin:3.1.0 requires Maven version 3.2.5 -> [Help 1]
So I had to install new version manually and finally got error that Java 11 is required.
Adding new dependency (maven-antrun-plugin) at this point doesn't look like good to me as it hides the real problem (Java version requirement).
Also I don't think there is a sense to have the test as it is already specified by maven compiler release setting.
So I suggest to open an issue about properties display in and discuss it there. 
For the issue this ticket should resolve I suggest removing maven-antrun-plugin and the test.

installer/pom.xml
```diff
@@ -248,4 +246,4 @@
             </plugin>
         </plugins>
     </reporting>
-</project>
+</project>
```
wwelling, line 249: Can we add the newline back to the end of this file?

api/pom.xml
```diff
@@ -21,6 +21,38 @@
 
     <build>
         <plugins>
+            <plugin>
```
wwelling, line 24: Please remove this plugin for future work under https://github.com/orgs/vivo-project/projects/6/views/2?pane=issue&itemId=21720002.

wwelling (changes requested):

api/pom.xml
```diff
@@ -21,6 +21,38 @@
...
+                                <echo message="[java.version]               ${java.version}" />
+                                <echo message="[java.specification.version] ${java.specification.version}" />
+                                <echo message="[maven.compiler.release]     ${maven.compiler.release}" />
+                                <echo message="###################################################################" />
+                                <echo message="###################################################################" />
+                            </target>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
```
chenejac, line 55:
```suggestion           
```

api/src/test/java/compilation/IsNotCompatibleWithJava8.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class IsNotCompatibleWithJava8 {
+/*
+ * This code segment is compatible with Java 11 and not compatible with java 8. It will abort the compilation if the Java context is version 8
+ */
+    @Test
+    public void test() {
+        String value = "This code is compiled in Java 11";
+        System.out.println(value.isBlank());
+    }
```
chenejac, line 15: I don't understand this. Maybe this might be useful - https://stackoverflow.com/questions/14572287/testing-different-java-versions-in-junit#answer-14573109

chenejac (changes requested):
@michel-heon please check my suggestions/comments

installer/pom.xml
michel-heon: Done

api/src/test/java/compilation/IsNotCompatibleWithJava8.java
michel-heon: This class uses a method (isBlank()) which is available only in JAVA 11. In java 8 the compilation sends an incompatibility message on the method.
The stackoverfow proposal is also very good. You just have to locate the right class to implement it. Do you have any suggestion ?

litvinovg: I don't think we need a test for that. I would suggest to remove it.

chenejac: The question is in which scenario someone can run tests using JDK 8, when we are requesting 11 in pom.xml file. It is of course possible, but completely out of VIVO installation/building procedure, should we take care in that case? Anyway, if some test is needed, it shouldn't just print some message, we need assertion (probably in this case assertFalse(value.isBlank()), but not sure what will happen in JDK 1.8 case).

PR #373, opened 2023-02-27
VIVO 3821: Add configuration to disable or enable Map of Science visualization.

kaladay:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues)**: [3821](https://github.com/vivo-project/VIVO/issues/3821)

Other Relevant Links (Mailing list discussion, related pull requests, etc.)
- For [Vivo PR 3841](https://github.com/vivo-project/VIVO/pull/3841)

Add an option to not show (disable) Map of Science Visualization link in the same manner as done with the Temporal Visualization link.

This is the Vitro part that adds the required variable to the theme builder code.

# What's new?
A new runtime property `visualization.mapOfScience` is added with the default set to `disabled`.

# How should this be tested?
Change the setting to enabled and disabled and check each theme to confirm behavior is consistent.

# Additional Notes:
I noticed that the Temporal Visualization setting is not applied in Vitro as it is done in Vivo. I added the Map of Science Visualization in the Vitro runtime configuration just like I did for the Vivo runtime configuration. This request is not about applying Temporal Visualization changes and so I left alone the (lack thereof) Temporal Visualization runtime configuration.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/home/src/main/resources/config/example.runtime.properties
+++ b/home/src/main/resources/config/example.runtime.properties
@@ -93,14 +93,6 @@ email.replyTo =
 # listview.usePreciseSubquery = true
 
 #
-# The Map of Science visualization is limited in the number of supported journals.
-# This will generally only be useful for institutions that specifically want this.
-# This is enabled by default.
-# Set this to either "disabled" to not utilize or "enabled" to utilize.
-#
-visualization.mapOfScience = enabled
-
-#
 # The email address of the root user for the VIVO application. The password
 # for this user is initially set to "rootPassword", but you will be asked to
 # change the password the first time you log in.
--- a/home/src/main/resources/config/example.runtime.properties
+++ b/home/src/main/resources/config/example.runtime.properties
@@ -95,10 +95,10 @@ email.replyTo =
 #
 # The Map of Science visualization is limited in the number of supported journals.
 # This will generally only be useful for institutions that specifically want this.
-# This is disabled by default.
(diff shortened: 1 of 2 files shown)


wwelling (dismissed):
LGTM. Only question I have is whether to enable by default?

wwelling (approved):

home/src/main/resources/config/example.runtime.properties
```diff
@@ -92,6 +92,14 @@ email.replyTo =
...
 # listview.usePreciseSubquery = true
 
+#
+# The Map of Science visualization is limited in the number of supported journals.
+# This will generally only be useful for institutions that specifically want this.
+# This is enabled by default.
+# Set this to either "disabled" to not utilize or "enabled" to utilize.
+#
+visualization.mapOfScience = enabled
+
```
chenejac, line 102: I think mapOfScience can't work in Vitro, therefore not sure there is a sense in this configuration in example.runtime.properties in Vitro project, I think we need that only in VIVO project (similar as visualization. temporal which I believe is only defined in VIVO runtime.properties). Later, in a separate issue, I hope we might even move the code from Vitro IndividualResponseBuilder to the VIVO codebase.

chenejac (changes requested):
I have posted one comment which we might discussed before merging this PR.

chenejac (approved):

PR #380, opened 2023-03-17
Added initial version of dump-restore script.

ivanmrsulja:
# What's new?
This functionality was only available through UI in admin panel, now, it is available through bash script so that it can be setup as cron job with minimal configuration.

# How should this be tested?
If you are willing to use the script the server has to be running. The CLI arguments are as follows:
```bash dumpRestore.sh <action> <model> <email> <password>```

Action can be: ```dump``` or ```restore```

Model can be: ```CONFIGURATION```, ```CONTENT``` or any other models that you have configured

Email and Password are your admin account email and password

In addition, you must configure some setup parameters. Because these parameters rarely change I figured that someone can just edit the first lines of the script so that the command doesn't get cluttered with a lot of arguments:

```
host="http://example.com/" # URL where the VIVO instance is hosted
purge="true"                 # Will the restoration process purge the models before restoring
restoration_files="."        # Directory containing the files used for restoration
dumped_files="."             # Directory containing the backed-up files
app_name="vitro_or_whatever"              # app-name parameter
```
# Interested parties
@chenejac @milospp

### DIFF
--- a/home/src/main/resources/scripts/dumpRestore.sh
+++ b/home/src/main/resources/scripts/dumpRestore.sh
@@ -15,8 +15,8 @@ session=$(mktemp -d)
 
 url="$host/$app_name/authenticate"
 
-loginName="email@email.com"
-loginPassword="password"
+loginName="$email"
+loginPassword="$password"
 loginForm="Log in"
 
 # Log in and get session cookies
--- a/home/src/main/resources/scripts/dumpRestore.sh
+++ b/home/src/main/resources/scripts/dumpRestore.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+action="$1"
+models="$2"
+email="$3"
+password="$4"
+
+host="http://localhost:8080" # URL where the VIVO instance is hosted
+purge="true"                 # Will the restoration process purge the models before restoring
+restoration_files="."        # Directory containing the files used for restoration
+dumped_files="."             # Directory containing the backed-up files
+app_name="vivo"              # app-name parameter
+
+session=$(mktemp -d)
+
+url="$host/$app_name/authenticate"
+
+loginName="email@email.com"
+loginPassword="password"
+loginForm="Log in"
+
+# Log in and get session cookies
+curl --cookie-jar "$session/cookies.txt" -d "loginName=$loginName" -d "loginPassword=$loginPassword" -d "loginForm=$loginForm" $url
+
+if [[ "$action" == "dump" ]]; then
+    echo "Starting dump..."
+    curl --cookie "$session/cookies.txt" "$host/$app_name/dumpRestore/dump/$models.nq?which=$models" -o "$dumped_files/$models.nq"
+    echo "Completed successfully."
+elif [[ "$action" == "restore" ]]; then
+    echo "Starting restoration process..."
+
+    url="$host/$app_name/dumpRestore/restore"
+    files=$(echo -n "$restoration_files/$models.nq")
+    params=$(echo -n "which=$models&purge=$purge")
+
+    curl --cookie "$session/cookies.txt" -X POST -F "sourceFile=@$files" "$url?$params" > /dev/null
+
+    if [[ $? -eq 0 ]]; then
+        echo "Restored successfully."
+    else
+        echo "Something went wrong."
+    fi
+fi
+
+rm -rf "$session"

### CODE REVIEW

**Comment by litvinovg** [comment]
I tried to dump data in linux environment, but it didn't work for me.
It prints that dump has been successfully created, but the file CONTENT.nq is empty
I suggest to add error checks and messages in case cookie wasn't returned and check return code of the dump request as well as the file size (not empty). 
Also it might be useful to use date in output file name.

curl 7.61.1 (x86_64-redhat-linux-gnu) libcurl/7.61.1 OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.6 libidn2/2.2.0 libpsl/0.20.2 (+libidn2/2.2.0) libssh/0.9.6/openssl/zlib nghttp2/1.33.0

File: home/src/main/resources/scripts/dumpRestore.sh [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+session=$(mktemp -d)
+
+url="$host/$app_name/authenticate"
+
+loginName="$email"
+loginPassword="$password"
+loginForm="Log in"
+
+# Log in and get session cookies
+curl --cookie-jar "$session/cookies.txt" -d "loginName=$loginName" -d "loginPassword=$loginPassword" -d "loginForm=$loginForm" $url
```
litvinovg, line 23: I think it should be --data-urlencode instead of -d as with -d "+" sign is interpreted as a space character

home/src/main/resources/scripts/dumpRestore.bat
```diff
@@ -0,0 +1,46 @@
...
+mkdir "%session%"
+
+set url=%host%/%app_name%/authenticate
+
+set loginName=%email%
+set loginPassword=%password%
+set loginForm=Log in
+
+:: Log in and get session cookies
+curl --cookie-jar "%session%\cookies.txt" --create-dirs -d "loginName=%loginName%" -d "loginPassword=%loginPassword%" -d "loginForm=%loginForm%" %url%
```
litvinovg, line 24: I think if you try to use login with + sign it wouldn't work here too.

home/src/main/resources/scripts/dumpRestore.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if [[ "$action" == "dump" ]]; then
+    echo "Starting dump..."
+    curl --cookie "$session/cookies.txt" "$host/$app_name/dumpRestore/dump/$models.nq?which=$models" -o "$dumped_files_path/$models.nq"
```
litvinovg, line 27: Check return code (should be 200, not 302)
If return code is 302 then most likely authorization failed.
I would also check size of returned file. It shouldn't be empty.

home/src/main/resources/scripts/dumpRestore.bat
```diff
@@ -0,0 +1,46 @@
...
+
+if "%action%" == "dump" (
+    echo Starting dump...
+    
+    curl --cookie "%session%\cookies.txt" "%host%/%app_name%/dumpRestore/dump/%models%.nq?which=%models%" -o "%dumped_files_path%\%models%.nq"
```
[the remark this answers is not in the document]
ivanmrsulja: I have added the necessary check, as you advised.

home/src/main/resources/scripts/dumpRestore.sh
ivanmrsulja: I have added the necessary check, as you advised.

ivanmrsulja:
I have improved the URL encoding and added necessary checks, as @litvinovg advised.

home/src/main/resources/scripts/dumpRestore.bat
litvinovg: It is still -d here. I haven't tested it on Windows, but maybe we should also use --data-urlencode here?

ivanmrsulja: You are right, I forgot to add it here for username and password. I will change it right now.

chenejac (changes requested):
@ivanmrsulja can we apply similar approach as Tomcat is using with setenv.sh (setenv.bat) file, meaning dumpRestore.bat and dumpRestore.sh might be renamed to vivo-admin.bat and vivo-admin.sh, and there might be additional files example.setenv.sh and example.setenv.bat? All scripts parameters might be specified in those setenv.sh/bat files except action. The files might stay at the **script** directory in the VIVO code base, but can be copied to VIVO_HOME/bin by using this file - https://github.com/vivo-project/Vitro/blob/main/home/src/main/assembly/home.xml

chenejac (approved):
@ivanmrsulja it works for me (Windows 10). I have tried to create dump of content graph and to restore.  Thanks for the contribution.

hauschke:
In case this is accepted, this solution should be included in  https://wiki.lyrasis.org/display/VIVODOC115x/Backup+and+Restore, right?

chenejac:
> In case this is accepted, this solution should be included in https://wiki.lyrasis.org/display/VIVODOC115x/Backup+and+Restore, right?

Right. This PR will be considered to be merged once jena is upgraded to 5 (or at least 4), hopefully in VIVO 1.16.x

bkampe (approved):
Installed and tested. Works like expected.

PR #381, opened 2023-04-17
Parameter substitutions

litvinovg:
Added parameter substitutions to improve procedure re-usability.
Some time ago we discussed that procedures should be reusable. So you could create one procedure and run it from another procedure. We created code to run procedures, but there was limitation: parameters names passed from caller to callable procedure should be the same, so you have to use different names everywhere. 
Now you can use parameter A in one procedure and rename it to B before sending to another procedure. The same with results if procedure returns C, but you need it to be named D, you can easily specify that.
The only thing required is to specify what to rename. 
Lines 15-27 [https://github.com/vivo-project/Vitro/blob/ed475f439f14a84ddb06abd63aa46788153f0b3[…]Substitution/parameter-substitution-call-exteranal-procedure.n3](https://github.com/vivo-project/Vitro/blob/ed475f439f14a84ddb06abd63aa46788153f0b35/api/src/test/resources/edu/cornell/mannlib/vitro/webapp/dynapi/components/ParameterSubstitution/parameter-substitution-call-exteranal-procedure.n3#L15)

# What's new?
Created component ParameterSubstitution to store substitution settings
Created component ParameterSubstitutor to perform parameters substitutions
Created ParameterSubstititionIntegrationTest and a few n3 configurations to use in it.

# How should this be tested?
* Run tests

# Interested parties
@chenejac


api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/components/ParameterSubstitutor.java
```diff
@@ -0,0 +1,52 @@
...
+                }
+            }
+        }
+    }
+
+    private static void substitute(Parameter source, Parameter target, DataStore dataStore) {
+        Data data = dataStore.getData(source.getName());
+        data.setParam(target);
+        dataStore.remove(source.getName());
+        dataStore.addData(target.getName(), data);
```
chenejac, line 37: what if target.getName() exists in data store for other steps? it will be overridden? can we make this substitution only local for an operationalStep, not impacting complete dataStore?

chenejac (changes requested):
@litvinovg I have posted just one comment, please check, it might be that you already addressed that case.

litvinovg:
Removed FORWARD and BACKWARD options to avoid case when parameter was removed but is required by other operations in procedure.
Added test for case when substitutions could create conflict with parameter names A -> B' And B -> A'. Fixed code for that case.

chenejac (approved):
@litvinovg looks well. Thanks

PR #383, opened 2023-04-24
Handle preferred language when creating new class and property groups

brianjlowe:
**[VIVO GitHub issue 3847](https://github.com/vivo-project/VIVO/issues/3847)**: 

When creating a new class or property group via the UI, uses current locale for label and public description instead of defaulting to en-US.

# What's new?
Propagates preferred languages list to temporary WebappDaoFactory with union models used to insert new groups.  Uses language-aware method for setting property group label and public description.

# How should this be tested?
Log in as admin.
Set locale to something other than en-US.
Create a new class group and property group.
Click through to groups you created.
Copy the value of the uri= parameter in the address bar.
Visit /entityEdit?uri= , pasting in the URI of the group.
View "Raw Statements with This Resource as Subject".
Confirm that label and public description (in case of property group) bear the appropriate language tag.
Switch do a different locale and edit groups.  Inspect triples again to confirm that values in the second language were added without affecting the first.

# Interested parties
@VIVO-project/vivo-committers


chenejac (approved):
@brianjlowe works as described.

litvinovg (approved):
Works well. Looks good to me.

PR #385, opened 2023-04-24
don't create add url if neither of 'select from existing' nor 'provid…

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3858)**

Prevents creation of button to add new object property if neither of 'select from existing' nor 'provide selection' options are set in  object property configuration

# How should this be tested?
* Remove options to select from existing and provide selection from object property
* Verify that add new object property button is not shown 
* Try the same for faux object property

# Interested parties
@chenejac


chenejac (approved):
@litvinovg well done

PR #387, opened 2023-04-25
fix: Escape data property value

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3864)**

Escapes returned value in defaultDataPropertyForm

# How should this be tested?

1. Create a data property (range: XMLLiteral, text editor: plaintext)
2. Create a value with " in the middle, save
3. Edit this value again

# Interested parties
@chenejac


chenejac (approved):
@litvinovg the statement editConfiguration.dataLiteralValuesAsString has been used in three more ftl files. Can you please check whether it should be also adjusted to display property with " in the middle?

PR #392, opened 2023-05-23
VIVO-3871 write TBox data in Add/Remove RDF with RDFService so as not to "shred" statements with blank nodes

brianjlowe:
**[VIVO-3871](https://github.com/vivo-project/VIVO/issues/3871)**:

Adds/removes TBox data on Add/Remove RDF page using the same technique already used for ABox data.  Because blank nodes are written to RDFService as part of a single ChangeSet, blank node statements remain intact.

# How should this be tested?
Upload dataset.ttl example file from linked issue.  Observe that triples are not shredded.  An easy way to do so is to issue this query from the SPARQL query page.  Leave syntax set to RDF/XML and you will see an intact nested superclass definition.
```
DESCRIBE ?x WHERE {
  ?x a owl:Class
  FILTER(REGEX(STR(?x), "vivo.mydomain.edu"))
}
```

# Interested parties


brianjlowe:
Will re-issue PR, because did not take into account updating of the in-memory cache.

brianjlowe:
Ignore previous PR completely, which I managed to screw up in multiple ways.  The problem looks much deeper and affects things like editing restrictions.  It seems that the bulk update handling for model adds/removes is no longer working properly, so will need to investigate that.

PR #396, opened 2023-06-12
skip request for graph uri from name

litvinovg:
Fixes DynapiModelFactory. Previously model uri was returned by model names map.
Dynamic api instances were changed to store uri some time ago, but DynapiModelFactory didn't reflect that.
This PR removes request to ModelNames and adds error log in case returned model is null and throws exception to stop execution.

# Interested parties
@chenejac @ivanmrsulja


ivanmrsulja (approved):
Looks good to me but we still need to check weather the reason that the response is always empty lies with an error in my n3 file or it is a new bug that we encountered.

PR #397, opened 2023-06-12
Use same logic to remove blank nodes from reasoner's internal model as used in RDFService

brianjlowe:
**[VIVO GitHub issue 3884](https://github.com/vivo-project/VIVO/issues/3884)**: 

Refactors the logic for treating removed blank nodes as variables from RDFServiceJena to JenaModelUtils where it can be used by the JFactTBoxReasoner.  This allows ontologies removed using Add/Remove RDF to properly update the reasoner and allow related inferences to be removed.  Adds some comments and a unit test.

# How should this be tested?
- Unit test should pass.
- For additional manual confirmation, run the steps in the linked issue to confirm that the bug can no longer be reproduced.

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/jena/JenaModelUtils.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/jena/JenaModelUtils.java
@@ -2,7 +2,6 @@
 
 package edu.cornell.mannlib.vitro.webapp.dao.jena;
 
-import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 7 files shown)


brianjlowe:
Hold off on this for a bit.  While the reasoner change is as expected, there is something here that caused the deletion of the asserted tbox to regress.  Investigating and will reopen.

brianjlowe:
I had introduced an error in the case where the blank-node-deletion code is called with a graph URI.  I simplified the existing complexity involving dataset / no dataset, graphURI / no graphURI and these methods now just take models.  The comments implied that the current situation was an optimization for SDB, but since we're removing SDB and these blank-node deletions should be infrequent things anyway, it seems like the simplification is preferable.

litvinovg (approved):
Looks good to me.

chenejac (approved):
@brianjlowe I have tested the scenario from the issue description. The issue is fixed. Also, I don't have complaints about the contributed code. Well done!!!

PR #399, opened 2023-06-15
Sparql select query fixes

litvinovg:
Fixed accidentally pasted text in model parameters (abox assertions graph uri).
Added check to prevent procedure initialization and if a model was configured as a parameter provided by SPARQL query.

# Interested parties


chenejac (approved):
@litvinovg thanks for this fix

PR #400, opened 2023-06-15
Fix for optional parameters

litvinovg:
* Renamed annotation with conflicting name
* Use empty json object if empty string was provided as input, treat missing node as invalid json in json converter.
* Added dynamic api ontology data property

# How to use
* To make parameter optional add default value, set parameter as internal and set parameter as optional.
        dynapi:defaultValue     "label" ;
        dynapi:isInternal       "true"^^xsd:boolean ;
        dynapi:isOptionalParam  "true"^^xsd:boolean ;

# Interested parties


ivanmrsulja (approved):
Great work!

chenejac (approved):

PR #402, opened 2023-06-16
Provide server url in DynamicAPIDocumentation

litvinovg:
Adds server url to Dynamic API documentation to fix swagger interface urls.

# What's new?
Added server url to open API. 
Added boolean variable in APIResponseMatcher to simplify updates of json responses.

# How should this be tested?
* Go to /dynapi-rpc-docs and /dynapi-rest-docs pages and try using endpoints via web interface

# Links 
[API Server and Base URL](https://swagger.io/docs/specification/api-host-and-base-path/) 
# Interested parties
@chenejac @ivanmrsulja @milospp @wwelling


chenejac (approved):
@litvinovg thanks. Works well, no comments for code improvements. Well done. I am behind http proxy, it also works in this case. Not sure, if there is a reverse proxy (nginx or apache httpd), but I suppose we can fix that later if there is a need.

PR #403, opened 2023-06-19
Added condition is not initialized by default

litvinovg:
Added ConditionIsNotInitializedByDefault to check if data is initialized with default value.

# Interested parties
@chenejac @milospp @ivanmrsulja


ivanmrsulja (approved):
Great work!

PR #410, opened 2023-06-27
Fallback to get user account from LoginStatusBean

litvinovg:
Added fallback to get user from LoginStatusBean

# How should this be tested?
* Log in into VIVO or Vitro as root
* Try access endpoints with swagger /dynapi-rest-docs
* Returned code shouldn't be 401

# Interested parties


chenejac (approved):
@litvinovg thanks for this. It works as described.

PR #411, opened 2023-06-27
Dynapi resource_id conversion fix

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3897)**

Fixes resource_id conversion
Changed report generator related procedure variable to resource_id.

# How should this be tested?
* Reproduce the problem you are fixing (if applicable)

# Interested parties

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/Endpoint.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dynapi/Endpoint.java
@@ -41,6 +41,9 @@ public abstract class Endpoint extends VitroHttpServlet {
         }
         DataStore dataStore = new DataStore();
         dataStore.setUser(user);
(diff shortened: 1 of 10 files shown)


ivanmrsulja (approved):
Works as expected, great job! From what I've seen in the code, whenever we want to use path variable to provide URI we have to specify parameter with exact name "resource_id"? Is that so or can we name it anything we want?

litvinovg:
> Works as expected, great job! From what I've seen in the code, whenever we want to use path variable to provide URI we have to specify parameter with exact name "resource_id"? Is that so or can we name it anything we want?

Yes. Also if you want to use other parameter name in sparql queries for example, you can create a procedure that has a resource_id variable, use it in endpoint configuration and forward the request to an internal procedure with procedure call operation.

chenejac (approved):
@litvinovg well done

PR #413, opened 2023-08-04
Custom REST actions fixes to support specified http methods and provide correct documentation

litvinovg:
Fixes Custom REST actions support. 
Implemented support of http methods in custom rest actions.

# How should this be tested?
Create a custom REST action with specific HTTP method. 
Verify that rest documentation is correct, try access endpoint in swagger interface

# Interested parties
@chenejac


chenejac (approved):
@litvinovg well done.

PR #419, opened 2023-09-13
Feature/setting locale url

ivanmrsulja:
Solves this [VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3906)

The link http://localhost/vivo/individual?uri=http://localhost/vivo/individual/pers_53a426a6-edc4-4f89-9bf6-c249fcf3b320&lang=pt_BR should open profile page for person with uri http://localhost/vivo/individual/pers_53a426a6-edc4-4f89-9bf6-c249fcf3b320 in the Portuguese language (IF PORTUGUESE LANGUAGE IS ADDED FIRST!).

# What's new?
Changed `languageSelector.ftl` so it parses URL as soon as the page is loaded to check if locale is specified. If it is, then the language list is checked to verify existance of such locale and, if successfull, updates the locale. Every successive change of locale will update the URL locale.

# How should this be tested?
Add some languages to VIVO first. Try to set any locale from URl (not just Potuguese :slightly_smiling_face:). Try to then manually change locale on that page. Everything should work as expected.

# Interested parties
@chenejac

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/selection/SelectedLocale.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/i18n/selection/SelectedLocale.java
@@ -4,6 +4,7 @@ package edu.cornell.mannlib.vitro.webapp.i18n.selection;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.IllformedLocaleException;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


brianjlowe (commented):
I would suggest taking a look at the getOverridingLocale() method in edu.cornell.mannlib.vitro.webapp.i18n.selection.SelectedLocale.  It seems it would be easier to add the lang parameter of the request to the existing local priority list (probably between "The forced Locale in the servlet context" and "The selected Locale in the session" as described in the Javadoc comment).  The lang parameter could also set the session locale to avoid needing to rewrite all subsequent URLs.  I think this feature could be implemented in a few lines of Java and would allow clients without Javascript (e.g. linked data requests) to also benefit from being able to specify the locale by parameter.

ivanmrsulja:
> I would suggest taking a look at the getOverridingLocale() method in edu.cornell.mannlib.vitro.webapp.i18n.selection.SelectedLocale. It seems it would be easier to add the lang parameter of the request to the existing locale priority list (probably between "The forced Locale in the servlet context" and "The selected Locale in the session" as described in the Javadoc comment). The lang parameter could also set the session locale to avoid needing to rewrite all subsequent URLs. I think this feature could be implemented in a few lines of Java and would allow clients without Javascript (e.g. linked data requests) to also benefit from being able to specify the locale by parameter.

I will take a look into that method tomorrow. I agree with you, solution on the backend seems better :slightly_smiling_face:

ivanmrsulja:
I have updated my PR as @brianjlowe suggested. Logic for updating URL lang parameter when clicking on an `<a>` tag is left on the client side because without it it would be really hard to detect whether someone selected a locale through URL or by clicking on an `<a>` tag.

PR #425, opened 2023-11-10
Updated argon2-jvm

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3915)**

Updates argon2-jvm dependency to support Mac M1

# How should this be tested?
Build and run VIVO on Mac M1

# Interested parties
@chenejac @milospp @gneissone

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -63,7 +63,7 @@
         <dependency>
             <groupId>de.mkammerer</groupId>
             <artifactId>argon2-jvm</artifactId>
-            <version>2.4</version>
+            <version>2.11</version>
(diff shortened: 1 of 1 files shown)


milospp (approved):
Works perfectly now on Mac M1 2020,

Btw I noticed that I was able to run VIVO with an old version of argon just by changing the JAVA_HOME version in catalina.sh to any that is based on x86_64

For example, if I change the JAVA_HOME variable to any of these Java versions with x86_64 tag, it also works fine:
image

chenejac (approved):

PR #427, opened 2023-11-22
Fix/captcha repeater vulnerability

ivanmrsulja:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues)**: [Mitigate vulnerability of Captcha feature](https://github.com/vivo-project/VIVO/issues/3918)
[VIVO PR](https://github.com/vivo-project/VIVO/pull/3920)

This PR mitigates vulnerability of captcha feature by introducing new methods of challenge generation: nanocaptcha and Google reCAPTCHA.

# What's new?
A CaptrchaServiceBean is created which can both provide and validate text-based challenges as well as provide validation services for Google reCaptcha through Google API. I also included a configuration for which captcha method is used so you can use both methods interchangeably (just change configuration options).

# How should this be tested?
Unit tests for CaptchaServiceBean are created. If you want to test manually, simply run the application and follow instructions on how to configure each captcha method (instructions are available in example.runtime.properties file).
Captcha is currently only used on `/contact` page. Keep in mind that you have to setup SMTP properties as well as default contact email for feedback form to be accessible.

# Interested parties
@litvinovg @chenejac @brianjlowe

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -71,6 +71,11 @@
             <version>1.5</version>
         </dependency>
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.25</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>fluent-hc</artifactId>
             <version>4.5.6</version>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -66,6 +66,11 @@
             <version>2.11</version>
         </dependency>
         <dependency>
+            <groupId>net.logicsquad</groupId>
+            <artifactId>nanocaptcha</artifactId>
+            <version>1.5</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>fluent-hc</artifactId>
             <version>4.5.6</version>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -73,7 +73,7 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
-            <version>1.7.25</version>
+            <version>1.7.36</version>
         </dependency>
         <dependency>
             <groupId>com.google.guava</groupId>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -76,6 +76,11 @@
             <version>1.7.25</version>
         </dependency>
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>30.1-jre</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>fluent-hc</artifactId>
             <version>4.5.6</version>
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/CaptchaServiceBean.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/CaptchaServiceBean.java
@@ -1,3 +1,5 @@
+/* $This file is distributed under the terms of the license in LICENSE$ */
+
 package edu.cornell.mannlib.vitro.webapp.beans;
 
 import java.awt.Color;
@@ -6,7 +8,6 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.security.SecureRandom;
 import java.util.Base64;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
@@ -16,8 +17,6 @@ import javax.imageio.ImageIO;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
-import edu.cornell.mannlib.vitro.webapp.config.ConfigurationProperties;
-import edu.cornell.mannlib.vitro.webapp.controller.VitroRequest;
 import net.logicsquad.nanocaptcha.image.ImageCaptcha;
 import net.logicsquad.nanocaptcha.image.backgrounds.GradiatedBackgroundProducer;
 import net.logicsquad.nanocaptcha.image.filter.FishEyeImageFilter;
@@ -58,13 +57,10 @@ public class CaptchaServiceBean {
      * Validates a reCAPTCHA response using Google's reCAPTCHA API.
      *
      * @param recaptchaResponse The reCAPTCHA response to validate.
-     * @param vreq              The VitroRequest associated with the validation.
+     * @param secretKey         The secret key used for Google ReCaptcha validation.
      * @return True if the reCAPTCHA response is valid, false otherwise.
      */
-    public static boolean validateReCaptcha(String recaptchaResponse, VitroRequest vreq) {
-        String secretKey =
-            Objects.requireNonNull(ConfigurationProperties.getBean(vreq).getProperty("recaptcha.secretKey"),
-                "You have to provide a secret key through configuration file.");
+    public static boolean validateReCaptcha(String recaptchaResponse, String secretKey) {
         String verificationUrl =
             "https://www.google.com/recaptcha/api/siteverify?secret=" + secretKey + "&response=" + recaptchaResponse;
 
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/CaptchaServiceBean.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/CaptchaServiceBean.java
@@ -8,6 +8,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.security.SecureRandom;
 import java.util.Base64;
+import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
@@ -17,6 +19,7 @@ import javax.imageio.ImageIO;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import edu.cornell.mannlib.vitro.webapp.config.ConfigurationProperties;
 import net.logicsquad.nanocaptcha.image.ImageCaptcha;
 import net.logicsquad.nanocaptcha.image.backgrounds.GradiatedBackgroundProducer;
 import net.logicsquad.nanocaptcha.image.filter.FishEyeImageFilter;
@@ -57,10 +60,12 @@ public class CaptchaServiceBean {
      * Validates a reCAPTCHA response using Google's reCAPTCHA API.
      *
      * @param recaptchaResponse The reCAPTCHA response to validate.
-     * @param secretKey         The secret key used for Google ReCaptcha validation.
      * @return True if the reCAPTCHA response is valid, false otherwise.
      */
-    public static boolean validateReCaptcha(String recaptchaResponse, String secretKey) {
+    public static boolean validateReCaptcha(String recaptchaResponse) {
+        String secretKey =
+            Objects.requireNonNull(ConfigurationProperties.getInstance().getProperty("recaptcha.secretKey"),
+                "You have to provide a secret key through configuration file.");
         String verificationUrl =
             "https://www.google.com/recaptcha/api/siteverify?secret=" + secretKey + "&response=" + recaptchaResponse;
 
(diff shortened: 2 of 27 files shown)


chenejac (changes requested):
@ivanmrsulja thanks for this contribution. Please check one my comment.

api/pom.xml
```diff
@@ -65,6 +65,16 @@
...
         </dependency>
+        <dependency>
+            <groupId>net.logicsquad</groupId>
+            <artifactId>nanocaptcha</artifactId>
+            <version>1.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.25</version>
```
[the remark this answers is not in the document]
ivanmrsulja: I have aligned the version as you suggested but if I try to move dependency to `dependencies/pom.xml` i get the following warning:
```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
SLF4J: Class path contains SLF4J bindings targeting slf4j-api versions 1.7.x or earlier.
SLF4J: Ignoring binding found at [jar:file:/home/ivanmrsulja/.m2/repository/org/slf4j/slf4j-log4j12/1.7.26/slf4j-log4j12-1.7.26.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See https://www.slf4j.org/codes.html#ignoredBindings for an explanation.
```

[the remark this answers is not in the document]
ivanmrsulja: Because of this, tests that rely on log parsing for functionality verification fail, because there is no log output without the necessary provider. It was the sole reason I added this dependency here which solved the issue. If you have another solution in mind I would be glad to switch to it :slightly_smiling_face: .

home/src/main/resources/rdf/i18n/de_DE/interface-i18n/firsttime/vitro_UiLabel.ttl
```diff
@@ -6269,3 +6269,43 @@ uil-data:no_individual_associated_with_id.Vitro
         uil:hasApp      "Vitro" ;
         uil:hasKey      "no_individual_associated_with_id" ;
         uil:hasPackage  "Vitro-languages" .
+
+uil-data:full_name_empty.Vitro
+        rdf:type         owl:NamedIndividual ;
+        rdf:type         uil:UILabel ;
+        rdfs:label       "Bitte geben Sie einen Wert im Feld Vollständiger Name ein."@de-DE ;
```
bkampe, line 6276:
```suggestion
        rdfs:label       "Bitte geben Sie ihren vollständigen Namen an."@de-DE ;
```

bkampe (changes requested):

home/src/main/resources/rdf/i18n/pt_BR/interface-i18n/firsttime/vitro_UiLabel.ttl
```diff
@@ -6269,3 +6269,43 @@ uil-data:no_individual_associated_with_id.Vitro
         uil:hasApp      "Vitro" ;
         uil:hasKey      "no_individual_associated_with_id" ;
         uil:hasPackage  "Vitro-languages" .
+
+uil-data:full_name_empty.Vitro
+        rdf:type         owl:NamedIndividual ;
+        rdf:type         uil:UILabel ;
+        rdfs:label       "Por favor, insira um valor no campo de nome completo."@pt-BR ;
```
salmjunior, line 6276:
```suggestion
        rdfs:label       "Por favor, insira um valor no campo, nome completo."@pt-BR ;
```

home/src/main/resources/rdf/i18n/pt_BR/interface-i18n/firsttime/vitro_UiLabel.ttl
```diff
@@ -6269,3 +6269,43 @@ uil-data:no_individual_associated_with_id.Vitro
...
+        rdf:type         uil:UILabel ;
+        rdfs:label       "Por favor, resolva o desafio CAPTCHA no campo de segurança fornecido."@pt-BR ;
+        uil:hasApp      "Vitro" ;
+        uil:hasKey      "captcha_user_sol_empty" ;
+        uil:hasPackage  "Vitro-languages" .
+
+uil-data:captcha_user_sol_invalid.Vitro
+        rdf:type         owl:NamedIndividual ;
+        rdf:type         uil:UILabel ;
+        rdfs:label       "Solução CAPTCHA inválida, por favor, tente novamente."@pt-BR ;
```
salmjunior, line 6308:
```suggestion
        rdfs:label       "A solução do CAPTCHA informada está inválida, por favor, tente novamente."@pt-BR ;
```

salmjunior (changes requested):

litvinovg (changes requested):
Thanks for the PR! 
Please take a look at some comments I did and translation suggestions.
I tested self hosted captcha. It works well.
Recaptcha V2 works.
Recaptcha V3 doesn't work (image below)
I didn't found any notes on what Recaptcha version should be supported. 
I think it should be documented or V3 fixed if that was expected to be working.
Screenshot from 2023-12-11 13-57-50

ivanmrsulja:
> Thanks for the PR! Please take a look at some comments I did and translation suggestions. I tested self hosted captcha. It works well. Recaptcha V2 works. Recaptcha V3 doesn't work (image below) I didn't found any notes on what Recaptcha version should be supported. I think it should be documented or V3 fixed if that was expected to be working. Screenshot from 2023-12-11 13-57-50

Yes, in the requirements, it was stated that there should be an "I'm not a robot field" which is reCAPTCHA v2 implementation. V3 runs in the background and it is a pure JS API that returns a score (based on number and speed of the interactions with the webpage). That score is the indicator whether additional measures should be taken for each interaction, it is a supplementary library. I did not plan to encorporate this functionality but it is open for discussion on the next meeting.

litvinovg (changes requested):
Please update comment in example.runtime.properties to mention that only re-captcha v2 keys are supported in this PR and VIVO PR as was discussed at the meeting on Tuesday.
I suggest to convert string values of captcha.implementation in Java code to static fields.
Also, the name for self hosted captcha can be better I think.
Maybe we could use implementation name, like nanocaptcha in case somebody in future decides to add another self hosted captcha implementation. Not sure it should be in upper case, lower case looks better to me.
@chenejac What is your opinion?

home/src/main/resources/config/example.runtime.properties
```diff
@@ -194,3 +194,11 @@ proxy.eligibleTypeList = http://www.w3.org/2002/07/owl#Thing
 #fileUpload.maxFileSize = 10485760
 #comma separated list of mime types allowed for upload
 #fileUpload.allowedMIMETypes = image/png, application/pdf
+
+# Captcha configuration. Available implementations are: DEFAULT (text-based) and RECAPTCHA
+# If captcha.implementation property is not provided, system will fall back to DEFAULT implementation
+# For RECAPTCHA method, you have to provide siteKey and secretKey.
+# More information on siteKey and secretKey is available on: https://www.google.com/recaptcha
+captcha.implementation = DEFAULT
```
chenejac, line 202: There should be captcha.enabled property with possible values true and false. The true is default one (if the property is not defined). 
Instead of DEFAULT we should use the nanocaptcha term for captcha.implementation property. Values for this property should be case insensitive (nanocaptcha and recaptcha can be also used). Also, there should be nanocaptcha.difficulty with two possible values easy and hard. Easy is the default value. 
```suggestion
captcha.implementation = nanocaptcha
nanocaptcha.difficulty = easy
```

chenejac (dismissed):
@ivanmrsulja I have added a comment which reflects committers' meeting discussion. Please, take a look.

home/src/main/resources/config/example.runtime.properties
ivanmrsulja: I have refactored my code and implemented everything you have suggested

api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/CaptchaServiceBean.java
```diff
@@ -0,0 +1,265 @@
...
+     */
+    public static String getCaptchaImpl() {
+        if (!Boolean.parseBoolean(ConfigurationProperties.getInstance().getProperty("captcha.enabled"))) {
+            return "NONE";
+        }
+
+        String captchaImpl =
+            ConfigurationProperties.getInstance().getProperty("captcha.implementation");
+        if (captchaImpl == null) {
+            captchaImpl = "NANOCAPTCHA";
```
litvinovg, line 175: Dragan wrote that values should be case insensitive, so this method is a good place to do case insensitive equality check.
I would refactor implementation options to an enumeration and return one of enumeration options here.

litvinovg (changes requested):
Please convert captcha implementation options to constants or enumerations throughout the PR. A few more comments in the code.

webapp/src/main/webapp/templates/freemarker/body/contactForm/contactForm-form.ftl
```diff
@@ -36,7 +36,7 @@
         </div>
 
 
-        <#if captchaToUse == "RECAPTCHA">
+        <#if captchaToUse == "RECAPTCHAv2">
```
chenejac, line 39: Any chance to make this case insensitive? Also, I suggest to use lowercase in example.runtime.properties (including text in comments, for instance recaptchav2 and nanocaptcha

chenejac (changes requested):

webapp/src/main/webapp/templates/freemarker/body/contactForm/contactForm-form.ftl
ivanmrsulja: It most certainly can be made case insensitive. I will switch to using lowercase as well.

litvinovg (changes requested):

api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/CaptchaServiceBean.java
```diff
@@ -0,0 +1,269 @@
...
+                break;
+            case NANOCAPTCHA:
+                Optional<CaptchaBundle> optionalChallenge = CaptchaServiceBean.getChallenge(challengeId);
+                if (optionalChallenge.isPresent() && optionalChallenge.get().getCode().equals(captchaInput)) {
+                    return true;
+                }
+                break;
+            case NONE:
+                return true;
+        }
```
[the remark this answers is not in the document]
ivanmrsulja: I can make a provider class for each implementation

api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/CaptchaServiceBean.java
```diff
@@ -0,0 +1,269 @@
...
+            return CaptchaImplementation.NONE;
+        }
+
+        String captchaImplSetting =
+            ConfigurationProperties.getInstance().getProperty("captcha.implementation");
+        if (Strings.isNullOrEmpty(captchaImplSetting)) {
+            captchaImplSetting = "NANOCAPTCHA";
+        }
+
+        return CaptchaImplementation.valueOf(captchaImplSetting.toUpperCase());
```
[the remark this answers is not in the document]
ivanmrsulja: You are right about misspelled values, but I think that it is more intuitive to firstly check all error logic as guard cases and then have the desired logic after that, looks more readable to me. I will refactor the code to support misspelled values and to address a suggestion below, and after that, if you still think this is unintuitive I can change it as well :smile:

api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/CaptchaServiceBean.java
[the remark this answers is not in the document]
ivanmrsulja: I have refactored code and made providers for each implementation. ServiceBean only delegates requests to providers and is responsible for challenge persistence (in-memory).

api/src/main/java/edu/cornell/mannlib/vitro/webapp/beans/CaptchaServiceBean.java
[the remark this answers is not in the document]
ivanmrsulja: This is improved to fall back to NANOCAPTCHA impl when misspelled configurations are provided.

home/src/main/resources/config/example.runtime.properties
```diff
@@ -194,3 +194,16 @@ proxy.eligibleTypeList = http://www.w3.org/2002/07/owl#Thing
...
+# Captcha configuration. Available implementations are: nanocaptcha (text-based) and recaptchav2
+# nanocaptcha is available in 2 difficulties (easy and hard)
+# If captcha.implementation property is not provided, system will fall back to nanocaptcha implementation
+# with easy diffuculty (if captcha is enabled)
+# For recaptchav2 method, you have to provide siteKey and secretKey.
+# More information on siteKey and secretKey is available on: https://www.google.com/recaptcha
+# Only recaptchav2 implementation is supported
+captcha.enabled = true
+captcha.implementation = nanocaptcha
+nanocaptcha.difficulty = hard
```
litvinovg, line 207:
```suggestion
nanocaptcha.difficulty = easy
```

litvinovg (commented):
Default difficulty should be easy.
I tested nanocaptcha and recaptchav2, works fine.

litvinovg (approved):

chenejac (approved):
@ivanmrsulja well done

PR #430, opened 2023-12-01
fix: added a check that the data set key is empty to prevent incorrec…

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3923)**

Avoid reuse of value sets by checking that the data set key is empty.

# How should this be tested?
* Test was created
* Try migration from ARM with multiple custom roles.

# Interested parties
@chenejac @matthiasluehr


chenejac (approved):
@litvinovg thanks for this. @matthiasluehr can you please confirm this is resolving the reported issue?

PR #435, opened 2023-12-21
fixed computational costs comparator, fixed test

litvinovg:
Fixed reversed order of computation costs, fixed test.

# How should this be tested?
Run test

# Interested parties
@chenejac


chenejac (approved):
@litvinovg thanks for this.

PR #445, opened 2024-03-01
provide full union model in authorization request for sparql checks

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3948)**

 Fixes failing sparql checks related to "suppress Display for object property in unrelated individuals", so that this not result in hidden object properties in related profiles.

# What's new?
Provided FULL_UNION model in HideFromDisplayByPolicyFilter to prevent SPARQL authorization checks from failing.

# How should this be tested?
* Reproduce the problem in the issue
* Test that the pull request fixes the issue

# Interested parties
@VIVO-project/vivo-committers


chenejac (approved):
@litvinovg works as described. 

Reproduced the issue before applying the fix from this PR, and tested it is fixed after applying this PR.

PR #456, opened 2024-04-02
Added search results sort fallback

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/3938)**

Added sort fallback option to provide an option to sort on multiple fields.

# What's new?
Added vitro-search:hasFallback object property to specify fallback sort options.
Defined raw name field and field sort options in search configuration.
Defined score field for default (relevance) sort option.
Used raw name sort options as fallback for language specific sort options.

# How should this be tested?
* Reproduce the problem in the issue
* Test that the pull request fixed the issue

# Interested parties
@VIVO-project/vivo-committers

### DIFF
--- a/home/src/main/resources/rdf/display/firsttime/search_individuals_vitro.n3
+++ b/home/src/main/resources/rdf/display/firsttime/search_individuals_vitro.n3
@@ -35,22 +35,45 @@
 :sort_title_desc  a                vitro-search:Sort ;
         vitro-search:order         20 ;
         vitro-search:sortField     :field_label_sort ;
+        vitro-search:hasFallback   :sort_name_raw_desc ;
         vitro-search:isAscending   false ;
         vitro-search:id            "titledesc" .
 
-:sort_title_asc  a                vitro-search:Sort ;
+:sort_title_asc  a                 vitro-search:Sort ;
         vitro-search:sortField     :field_label_sort ;
         vitro-search:isAscending   true ;
+        vitro-search:hasFallback   :sort_name_raw_asc ;
         vitro-search:order         30 ;
         vitro-search:id            "titleasc" .
 
+:sort_name_raw_desc  a             vitro-search:Sort ;
+        vitro-search:order         40 ;
+        vitro-search:sortField     :field_name_raw ;
+        vitro-search:isAscending   false ;
+        vitro-search:id            "name_raw_desc" .
+
+:sort_name_raw_asc  a                vitro-search:Sort ;
+        vitro-search:sortField     :field_name_raw ;
+        vitro-search:isAscending   true ;
+        vitro-search:order         50 ;
+        vitro-search:id            "name_raw_asc" .
+
 :sort_by_relevance a                vitro-search:Sort ;
+        vitro-search:sortField     :field_score ;
         vitro-search:id            "relevance" .
+
+:field_score
+        a                          vitro-search:SearchField ;
+        vitro-search:indexField    "score" .
 
 :field_label_sort
         a                          vitro-search:SearchField ;
         vitro-search:isLanguageSpecific true ;
         vitro-search:indexField    "_label_sort" .
+
+:field_name_raw
+        a                          vitro-search:SearchField ;
+        vitro-search:indexField    "nameRaw" .
 
 :field_category
         a                          vitro-search:SearchField ;
--- a/home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
+++ b/home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
@@ -12,3 +12,7 @@
 :field_type rdfs:label "Typ"@de-DE .
 :filter_querytext rdfs:label "Text"@de-DE .
 :field_querytext rdfs:label "Standardfeld"@de-DE .
+:sort_name_raw_desc rdfs:label "Roher Titel Z-A"@de-DE .
+:sort_name_raw_asc rdfs:label "Roher Titel A-Z"@de-DE .
+:field_score rdfs:label "Relevanzfeld"@de-DE .
+:field_name_raw rdfs:label "Rohes Titelfeld"@de-DE .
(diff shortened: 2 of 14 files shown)


chenejac (commented):
@milospp, @litvinovg I can't reproduce the issue before application of this PR. 
I have tried to turn off multilingual UI in the runtime.properties, but it is still sorting well.

chenejac (changes requested):
@litvinovg thanks for the PR. 

I have reproduced the issue before applying this fix. The PR is fixing the issue. However, there are five options if multilinguality of UI is turned off:

Sort by Relevance
Sort by Title Z-A
Sort by Title A-Z
Sort by Raw title Z-A
Sort by Raw title A-Z

The second and fourth options, as well as third and fifth options are working the same, but the last two options should be hidden (not to confuse the user).

litvinovg:
> The second and fourth options, as well as third and fifth options are working the same, but the last two options should be hidden (not to confuse the user).
@chenejac I addressed that in recent commit, please check

home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
```diff
@@ -12,3 +12,7 @@
 :field_type rdfs:label "Typ"@de-DE .
 :filter_querytext rdfs:label "Text"@de-DE .
 :field_querytext rdfs:label "Standardfeld"@de-DE .
+:sort_name_raw_desc rdfs:label "Roher Titel Z-A"@de-DE .
+:sort_name_raw_asc rdfs:label "Roher Titel A-Z"@de-DE .
```
chenejac, line 16: vitro-search:display property for sort_name_raw_desc is missing, meaning it is false by default, meaning with the current configuration these labels are not needed. I think if someone decide to turn it on (set display to true), he/she should add labels for languages he/she needs. Not sure we need this in the codebase? The comment is the same for all other languages for those two labels.

home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
```diff
@@ -12,3 +12,7 @@
...
+:field_score rdfs:label "Relevanzfeld"@de-DE .
+:field_name_raw rdfs:label "Rohes Titelfeld"@de-DE .
```
chenejac, line 18: Where those two labels are used?

home/src/main/resources/rdf/display/firsttime/search_individuals_vitro.n3
```diff
@@ -35,23 +35,49 @@
...
+        vitro-search:order         40 ;
+        vitro-search:sortField     :field_name_raw ;
+        vitro-search:isAscending   false ;
+        vitro-search:id            "name_raw_desc" .
+
+:sort_name_raw_asc  a                vitro-search:Sort ;
+        vitro-search:sortField     :field_name_raw ;
+        vitro-search:isAscending   true ;
+        vitro-search:order         50 ;
+        vitro-search:id            "name_raw_asc" .
```
chenejac, line 61: it would be nice if order of properties is the same in all classes

chenejac (changes requested):
@litvinovg thanks for the improvements. It works nice, I have tested the feature with languageFilters and without. Please take a look into my comments.

home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
litvinovg: In first version of extended search configuration was stored in content store, so users could navigate through configuration. Labels will be useful when we create an option to navigate over search settings in display graph.

home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
litvinovg: Those labels were created for the same reason as described above. For future work with search settings in web interface.

home/src/main/resources/rdf/display/firsttime/search_individuals_vitro.n3
litvinovg: Fixed

home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
chenejac: ok, let's keep these labels. I will organize validation of messages by native speakers

chenejac (approved):

home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
```diff
@@ -12,3 +12,7 @@
 :field_type rdfs:label "Typ"@de-DE .
 :filter_querytext rdfs:label "Text"@de-DE .
 :field_querytext rdfs:label "Standardfeld"@de-DE .
+:sort_name_raw_desc rdfs:label "Roher Titel Z-A"@de-DE .
```
bkampe, line 15:
```suggestion
:sort_name_raw_desc rdfs:label "Titel Z-A"@de-DE .
```

home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
bkampe, line 16:
```suggestion
:sort_name_raw_asc rdfs:label "Titel A-Z"@de-DE .
```

home/src/main/resources/rdf/i18n/de_DE/display/firsttime/search_individuals_vitro.n3
bkampe, line 18:
```suggestion
:field_name_raw rdfs:label "Titelfeld"@de-DE .
```

bkampe (changes requested):
"Raw" is not used that way in german, I would suggest to simply remove it.

chenejac (approved):
@litvinovg this looks ok to me.

milospp (approved):
@litvinovg Thanks for the fix, this is working for me.

PR #460, opened 2024-04-09
Wilma Responsive redesign

milospp:
**[GitHub issue 3910](https://github.com/vivo-project/VIVO/issues/3910)**
**[GitHub PR VIVO 3972](https://github.com/vivo-project/VIVO/pull/3972)**

Added bootstrap (only grid style) CSS and modified Wilma theme to leverage its responsive design feature and enhance the overall layout.

# What's new?
Added bootstrap v5
Refactored nav tabs in Person, Org Unit page to use bootstrap component instead of jQuery
Refactored files locations

# How should this be tested?
Open every page and test by resizing up to around 400px in width (Mobile device toolbar can be used in google chrome to simulate phone display)

# Additional Notes:
Here is the video example

- The Left window is VIVO with old Wilma theme
- The middle window is a new, updated Wilma theme
- The right window is Tenderfoot theme (just for comparison)

[Video Example](https://youtu.be/9B0_cevrEeE)
https://youtu.be/9B0_cevrEeE


webapp/src/main/webapp/js/individual/propertyGroupControls.js
```diff
@@ -163,27 +134,23 @@ $(document).ready(function(){
...
-                    // select the stored tab
-                    $("li[groupName='" + groupName + "']").removeClass("nonSelectedGroupTab clickable");
-                    $("li[groupName='" + groupName + "']").addClass("selectedGroupTab clickable");
-                    // hide the first tab section
-                    $('section.property-group:visible').hide();
 
                     if ( groupName == "viewAll" ) {
-                        processViewAllTab();
-                    }
+                        console.log("View all")
```
balmas, line 141: suggest removing console debugging print

balmas (commented):
The code and responsiveness looks good to me but see my comment in https://github.com/vivo-project/VIVO/pull/3972 about a problem loading the temporal google charts

balmas (approved):
Given that the chart issue is already fixed in main and should be resolved in the merge, I think that this looks good

litvinovg (commented):
I suggest to use webjar dependency https://github.com/vivo-project/Vitro/pull/467
Also I think boostrap3 files shouldn't be moved as they are out of scope for this PR (Wilma didn't use boostrap3 before).

webapp/src/main/webapp/js/individual/propertyGroupControls.js
milospp: Thanks, I removed it.

milospp:
> I suggest to use webjar dependency #467 Also I think boostrap3 files shouldn't be moved as they are out of scope for this PR (Wilma didn't use boostrap3 before).

I agree that Bootstrap 3 location refactor is not related to the Wilma theme directly. I can revert those changes, but would it be okay to have Bootstrap 5 on this location:
Vitro/webapp/src/main/webapp/bootstrap-5.3.2
and Bootstrap 3 on this:
Vitro/webapp/src/main/webapp/js/bootstrap

(Bootstrap 3 files are stored inside js subfolder)

litvinovg:
> > I suggest to use webjar dependency #467 Also I think boostrap3 files shouldn't be moved as they are out of scope for this PR (Wilma didn't use boostrap3 before).
> 
> I agree that Bootstrap 3 location refactor is not related to the Wilma theme directly. I can revert those changes, but would it be okay to have Bootstrap 5 on this location: Vitro/webapp/src/main/webapp/bootstrap-5.3.2 and Bootstrap 3 on this: Vitro/webapp/src/main/webapp/js/bootstrap
> 
> (Bootstrap 3 files are stored inside js subfolder)

With webjar there would be no need to upload bootstrap js/css assets into the project on each update and the path would /webjars/bootstrap/ (without version), which is important for customized themes update as js/css paths wouldn't change on next update.
Also all assets would be in one place, not like [this](https://github.com/vivo-project/VIVO/pull/3972/files#diff-a9b05c73ee4615105f6a46ba4cc1fafb5c7bca3bf877c103efb5ada33441d1d3) one.

balmas:
@milospp is this one ready for re-review now?

gneissone (approved):
Looks great, thank you.

litvinovg (changes requested):
This set of PRs not only modifies Wilma theme to have responsive design, but also introduces global dependency for shared components on specific version of Bootstrap which can't safely be done retrospectively.
So it breaks backward compatibility for customized themes
https://wiki.lyrasis.org/display/VIVODOC115x/Creating+a+custom+theme
I think we should find a better way to do that.

PR #469, opened 2024-06-10
Small mapping update and response parsing fix.

ivanmrsulja:
Updates current ES 6.x integration to 8.x.

# What's new?
Changes in ResponseParser and ES documentation on the first draft.

Example:
* Changed `src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ResponseParser.java` to be in line with current ES API
* Updated `src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/Elasticsearch_notes_on_the_first_draft.md` with new mapping
* Updated example.applicationSetup.n3 to show ES setup example

# How should this be tested?
Common manual tests should be performed.

# Interested parties
@chenejac 

# Reviewers' expertise

1. Java
2. Elasticsearch 8.*

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/Elasticsearch_notes_on_the_first_draft.md
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/Elasticsearch_notes_on_the_first_draft.md
@@ -65,57 +65,79 @@ git clone -b feature/elasticsearchExperiments https://github.com/j2blake/Vitro.g
 ```
 curl -X PUT "localhost:9200/vivo?pretty" -H 'Content-Type: application/json' -d'
 {
-  "mappings": {
-    "_doc": { 
-      "properties": { 
-        "ALLTEXT": { 
-          "type": "text",
-          "analyzer": "english"
-        }, 
-        "ALLTEXTUNSTEMMED": { 
-          "type": "text",
-          "analyzer": "standard"
-        }, 
-        "DocId": {
-          "type": "keyword"  
-        }, 
-        "classgroup": {
-          "type": "keyword"  
-        }, 
-        "type": {
-          "type": "keyword"  
-        }, 
-        "mostSpecificTypeURIs": {
-          "type": "keyword"  
-        }, 
-        "indexedTime": { 
-          "type": "long" 
-        },
-        "nameRaw": { 
-          "type": "keyword" 
-        },
-        "URI": { 
-          "type": "keyword" 
-        },
-        "THUMBNAIL": { 
-          "type": "integer" 
-        },
-        "THUMBNAIL_URL": { 
-          "type": "keyword" 
-        },
-        "nameLowercaseSingleValued": {
-          "type": "text",
-          "analyzer": "standard",
-          "fielddata": "true"
-        },
-        "BETA" : {
-          "type" : "float"
+  "settings": {
+    "index": {
+      "analysis": {
+        "analyzer": {
+          "default": {
+            "type": "english"
+          }
         }
       }
     }
   },
-  "query": {
-    "default_field": "ALLTEXT"
+  "mappings": {
+    "dynamic_templates": [
+      {
+        "field_sort_template": {
+          "match": "*_label_sort",
+          "mapping": {
+            "type": "text",
+            "fields": {
+              "keyword": {
+                "type": "keyword"
+              }
+            },
+            "fielddata": true
+          }
+        }
+      }
+    ],
+    "properties": { 
+      "ALLTEXT": { 
+        "type": "text",
+        "analyzer": "english"
+      }, 
+      "ALLTEXTUNSTEMMED": { 
+        "type": "text",
+        "analyzer": "standard"
+      }, 
+      "DocId": {
+        "type": "keyword"  
+      }, 
+      "classgroup": {
+        "type": "keyword"  
+      }, 
+      "type": {
+        "type": "keyword"  
+      }, 
+      "mostSpecificTypeURIs": {
+        "type": "keyword"  
+      }, 
+      "indexedTime": { 
+        "type": "long" 
+      },
+      "nameRaw": { 
+        "type": "keyword" 
+      },
+      "URI": { 
+        "type": "keyword" 
+      },
+      "THUMBNAIL": { 
+        "type": "integer" 
+      },
+      "THUMBNAIL_URL": { 
+        "type": "keyword" 
+      },
+      "nameLowercaseSingleValued": {
+        "type": "text",
+        "analyzer": "standard",
+        "fielddata": true
+      },
+      "BETA" : {
+        "type" : "float"
+      }
+    }
   }
 }
 '
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ResponseParser.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ResponseParser.java
@@ -98,32 +98,36 @@ class ResponseParser {
         highlightingMap = new HashMap<>();
 
         @SuppressWarnings("unchecked")
-        Map<String, Object> uberHits = (Map<String, Object>) responseMap
-                .get("hits");
+        Map<String, Object> uberHits = (Map<String, Object>) responseMap.get("hits");
         if (uberHits == null) {
-            log.warn("Didn't find a 'hits' field " + "in the query response: "
-                    + responseMap);
+            log.warn("Didn't find a 'hits' field in the query response: " + responseMap);
             return;
         }
 
-        Integer total = (Integer) uberHits.get("total");
+        // Updated handling of the 'total' field
+        @SuppressWarnings("unchecked")
+        Map<String, Object> totalMap = (Map<String, Object>) uberHits.get("total");
+        if (totalMap == null) {
+            log.warn("Didn't find a 'hits.total' field in the query response: " + responseMap);
+            return;
+        }
+        Integer total = ((Number) totalMap.get("value")).intValue(); // Extract the integer value
+
         if (total == null) {
-            log.warn("Didn't find a 'hits.total' field "
-                    + "in the query response: " + responseMap);
+            log.warn("Didn't find a 'hits.total.value' field in the query response: " + responseMap);
             return;
         }
 
         @SuppressWarnings("unchecked")
-        List<Map<String, Object>> hits = (List<Map<String, Object>>) uberHits
-                .get("hits");
+        List<Map<String, Object>> hits = (List<Map<String, Object>>) uberHits.get("hits");
         if (hits == null) {
-            log.warn("Didn't find a 'hits.hits' field "
-                    + "in the query response: " + responseMap);
+            log.warn("Didn't find a 'hits.hits' field in the query response: " + responseMap);
             return;
         }
 
         parseDocuments(hits);
     }
+
 
     private void parseDocuments(List<Map<String, Object>> hits) {
         for (Map<String, Object> hit : hits) {
--- a/home/src/main/resources/config/example.applicationSetup.n3
+++ b/home/src/main/resources/config/example.applicationSetup.n3
@@ -69,13 +69,22 @@
 #    and more rigorous life-cycle checking.
 #
 
+#:instrumentedSearchEngineWrapper
+#    a   vitroWebapp:searchengine.InstrumentedSearchEngineWrapper ,
+#        vitroWebapp:modules.searchEngine.SearchEngine ;
+#    :wraps :solrSearchEngine .
+
+#:solrSearchEngine
+#    a   vitroWebapp:searchengine.solr.SolrSearchEngine ,
+#        vitroWebapp:modules.searchEngine.SearchEngine .
+
 :instrumentedSearchEngineWrapper
     a   vitroWebapp:searchengine.InstrumentedSearchEngineWrapper ,
         vitroWebapp:modules.searchEngine.SearchEngine ;
-    :wraps :solrSearchEngine .
+    :wraps :elasticSearchEngine .
 
-:solrSearchEngine
-    a   vitroWebapp:searchengine.solr.SolrSearchEngine ,
+:elasticSearchEngine
+    a   vitroWebapp:searchengine.elasticsearch.ElasticSearchEngine ,
         vitroWebapp:modules.searchEngine.SearchEngine .
 
 # ----------------------------


chenejac:
The following features should be tested:

- search form - searching, filtering and sorting (changing localization)
- list of research, persons, org units - alphabetical index (changing localization)
- lookup/autocompletion at some forms - for instance adding author to a publication (changing localization)
- visualizations - map of science, collaboration network

chenejac:
@ivanmrsulja please create a VIVO PR with updated example.runtime.properties. Also, please move JSON configuration into [vivo-es project](https://github.com/vivo-project/vivo-es). Add in the vivo-es project a Docker file, and update README file to explain how ES should be run.

chenejac (approved):
@ivanmrsulja basic VIVO search functionalities works for me. I didn't review the code. Instructions from the PR description about setup of the elasticsearch index might be replaced with a pointer to the vivo-es Readme file.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/base/SearchEngineUtil.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class SearchEngineUtil {
+
+    private static final Log log = LogFactory.getLog(SearchEngineUtil.class);
+
+    public static String getSearchEngineURLProperty() {
+        ConfigurationProperties config = ConfigurationProperties.getInstance();
+        if (config.getProperty("vitro.local.searchengine.url", "").isEmpty()) {
```
wwelling, line 13: If `ConfigurationProperties.getInstance()` returns the DummyConfigurationProperties [getProperty](https://github.com/vivo-project/Vitro/blob/main/api/src/main/java/edu/cornell/mannlib/vitro/webapp/config/DummyConfigurationProperties.java#L24) will return null and isEmpty call will throw a null pointer exception.

wwelling: For the public static method to be safely used, please add appropriate null checks. And if possible, add JavaDocs describing that the method may return null.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/CustomQueryBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+import co.elastic.clients.elasticsearch._types.query_dsl.PrefixQuery;
+import co.elastic.clients.elasticsearch._types.query_dsl.Query;
+import co.elastic.clients.elasticsearch._types.query_dsl.RangeQuery;
+import co.elastic.clients.elasticsearch._types.query_dsl.WildcardQuery;
+
+public class CustomQueryBuilder {
+
+    private static final String MAX_FUZZY_EDITS = "2";
+
+    public static Query buildQuery(SearchType queryType, String field, String value) {
```
wwelling, line 16: Please add a private default constructor to avoid misuse of this factory builder class.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/base/SearchEngineUtil.java
ivanmrsulja: Good find! I did as you proposed. Also, I included `@Nullable` annotation so it is clear that this method can return `null`.

litvinovg:
It seems the mapping in the PR description is not the same as the mapping in https://github.com/ivanmrsulja/vivo-es/blob/main/index-config.json 
Also if search string is empty then there are not search result on search page, also no filters are visible, that is different from current behavior with Solr based search indexes.

litvinovg:
Thanks a lot! 
I just found one more issue while trying date range slider (similar to [this](https://sah.tib.eu/travel_sketches?filter_range_dates=1521-12-31T23%3A06%3A32.000Z+1929-12-31T23%3A00%3A00.000Z&filters_label_regex=label_regex%3A&hitsPerPage=30&lang=de-DE)) . Could you take a look?
[error.log](https://github.com/user-attachments/files/20704837/error.log)
Example filter to reproduce (remove .txt suffix and put into rdf/display/firsttime directory )[time_period_example.n3.txt](https://github.com/user-attachments/files/20705830/time_period_example.n3.txt)

ivanmrsulja:
> Thanks a lot! I just found one more issue while trying date range slider (similar to [this](https://sah.tib.eu/travel_sketches?filter_range_dates=1521-12-31T23%3A06%3A32.000Z+1929-12-31T23%3A00%3A00.000Z&filters_label_regex=label_regex%3A&hitsPerPage=30&lang=de-DE)) . Could you take a look? [error.log](https://github.com/user-attachments/files/20704837/error.log) Example filter to reproduce (remove .txt suffix and put into rdf/display/firsttime directory )[time_period_example.n3.txt](https://github.com/user-attachments/files/20705830/time_period_example.n3.txt)

Should be fixed now, please test it out when you have time :smile:

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ESQuery.java
```diff
@@ -34,9 +38,17 @@ public ESQuery(String baseUrl) {
 
     public SearchResponse query(SearchQuery query)
             throws SearchEngineException {
-        String queryString = new QueryConverter(query).asString();
+        QueryParser parser = new QueryParser("defaultField", new StandardAnalyzer());
+        boolean treatAsLuceneQuery = true;
+        try {
+            parser.parse(query.getQuery());
+        } catch (ParseException e) {
+            treatAsLuceneQuery = false;
```
litvinovg, line 46: Isn't the Lucene query format used for all current search queries in VIVO?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ExpressionTransformer.java
```diff
@@ -0,0 +1,341 @@
+package edu.cornell.mannlib.vitro.webapp.searchengine.elasticsearch;
```
chenejac, line 1: This class (and linked classes CustomQueryBuilder and SearchType) make this PR a little bit more complex. If I understood well using query_string in ElasticSearch is possible, meaning to pass query to ElasticSearch and to wait for a response, but it is less powerful than Ivan's approach and it is not a good practice especially for search boxes (check the first warning box at this [link](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-query-string-query), and also this [discussion](https://discuss.elastic.co/t/should-i-use-query-string-in-production/48718)). 

I think regarding some advance features of query syntax of ElasticSearch query language over Lucene query_string, we might conclude it is not crucial for us at the moment, because we are expecting end-users wouldn't use advance query syntax elements in the search box, they will use only content they are looking for (meaning we need keywords searching). 

However, the question is whether it is a good practice to use query_string via search box due to the following limitations:

- While versatile, the query is strict and returns an error if the query string includes any invalid syntax (due to some brackets, columns, and other elements which might be present in title of work someone is looking for) - [source](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-query-string-query). This maybe might be fixed by using `lenient` flag. 
- A malicious bot could inject special Lucene syntax like wildcards, range queries, or even malformed expressions, leading to performance degradation (e.g., complex regex/wildcard queries), etc. 

Can we discuss here advantages and disadvantages of query_string and ElasticSearch DSL query? 
What about [simple_query_string](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-simple-query-string-query)? It might be more safe for us. 
Moreover, I am wondering whether the listed issues above are also present when we are using Solr?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/QueryConverter.java
```diff
@@ -140,25 +197,25 @@ public QueryStringMap(String queryString) {
 
         /**
          * This is a kluge, but perhaps it will work for now.
-         * 
+         * <p>
```
chenejac, line 200: Please remove those p tags if they are not needed.

chenejac (changes requested):
@ivanmrsulja please check my comments.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ESQuery.java
ivanmrsulja: Unfortunately not, if you add a query like `:something"(` that is regarded as an invalid Lucene query, Solr handles this implicitly by treating everything as full text.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ExpressionTransformer.java
ivanmrsulja: Thanks for the detailed comment! You're absolutely right to bring up the trade-offs between query_string, simple_query_string, and our current custom query parser based approach. Let me add some thoughts on why I’ve stuck with the custom builder and why switching to `query_string` (or even `simple_query_string`) might not be the best path forward in our case:

- As Georgy mentioned, users often submit queries like `Deep Learning: A Survey (2023)`, with `query_string`, these inputs must be perfectly formed Lucene syntax which users won’t know. Even a missing quote or special character can cause parsing failures or incorrect behavior. While the `lenient` flag can help, it can also mask deeper issues and result in confusing results, often returning 0 results because of parse failure or malformed query (e.g. `:something"(`).

- Using `query_string` directly opens the door to Lucene syntax injection. Malicious users or bots could send wildcard-heavy, deeply nested, or regex-based queries that can degrade search performance or cause errors. Our current approach allows us to filter, escape, or block these patterns early before hitting Elasticsearch.

- Elasticsearch’s query DSL gives us the ability to define clear search logic with must, should, boost, and filter. We can control how fields like `title` and `author` contribute to scoring, or apply different analyzers. `query_string` flattens this control, and it becomes harder to evolve the search experience in the future.

- Since we must support both structured and free-text queries in the same engine endpoint (I have no known way to differentiate where the query came from, and not all structured parts are in filters), a naive switch to `query_string` wouldn’t improve our situation. It would require extra parsing or escaping on our side anyway. Our current query builder handles both gracefully and consistently, keeping our logic centralized and testable. `simple_query_string` also doesn’t support more advanced query structures or field-level boosting (which we might want to add in the future). It's a safer subset of `query_string`, but I think it will not be flexible enough even for our current feature set.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/QueryConverter.java
ivanmrsulja: This is a codestyle convention, they are needed if you want a blank line in the intellisense or geerated documentation.

image

litvinovg:
I tried to test it once again.
Search query input example  andrew:"bogart("
[error-01.08.25.log](https://github.com/user-attachments/files/21543237/error-01.08.25.log)

ivanmrsulja:
> I don't see filters if I open search page.

I fixed the empty search field edgecase :slightly_smiling_face:

ivanmrsulja:
> The edge case is fixed. Why should it only be applied to the /search page? Is it possible to add tests and clean up the code?

I updated the code as you suggested. Also, I cleaned up stale, commented-out code and dependencies and added tests for utility classes.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/search/controller/PagedSearchController.java
```diff
@@ -203,6 +204,14 @@ public static ResponseValues process(VitroRequest vreq, Map<String, List<String>
             SearchResponse response = null;
 
             try {
+                if (((HttpServletRequest) vreq.getRequest()).getRequestURI().endsWith("/search")) {
```
litvinovg, line 207: I think we should avoid conditionals based on servlet name.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/base/BaseSearchQuery.java
```diff
@@ -29,6 +29,7 @@ public class BaseSearchQuery implements SearchQuery {
 	private int facetMinCount = -1;
     private boolean facetTextToCompareIgnoreCase;
     private String facetContainsText;
+	private boolean isSimpleSearchQuery = false;
```
litvinovg, line 32: name should be aligned with setter/getter names

api/src/main/java/edu/cornell/mannlib/vitro/webapp/search/controller/PagedSearchController.java
ivanmrsulja: Unfortunately this was the only way I could differentiate from which actual page is this particular servlet called, furthermore, only search page requests require plain text query interpretation. Do you have another suggestion on how I can implement this?

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ESAdder.java
```diff
@@ -75,15 +92,51 @@ private Map<String, List<Object>> convertDocToMap(SearchInputDocument doc) {
...
                     + URLEncoder.encode(docId, "UTF8");
-            Response response = Request.Put(url)
-                    .bodyString(json, ContentType.APPLICATION_JSON).execute();
-            log.debug("Response from Elasticsearch: "
-                    + response.returnContent().asString());
+            HttpClient httpClient;
+            if (baseUrl.startsWith("https")) {
+                httpClient = ESHttpsBasicClientFactory.getHttpClient();
+            } else {
+                httpClient = HttpClientFactory.getHttpClient();
```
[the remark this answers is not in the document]
ivanmrsulja: You are right. I encapsulated everything into one shared factory. Great suggestion!

litvinovg:
It seems ssl connection doesn't work.
[error.log](https://github.com/user-attachments/files/23656483/error.log)

litvinovg:
Thank you for the fix.
What I also noticed is that date time filters work with OpenSearch, but do not work with ElasticSearch.
@ivanmrsulja could you check it?

PR #482, opened 2024-12-12
VIVO-4030

litvinovg:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/4030)**
[Related VIVO PR](https://github.com/vivo-project/VIVO/pull/4054)

Created policy data sets for publish operations on properties by users with public role.
Created policy data sets for publish operations on classes by users with public role.
Removed exceptional case related to publish operation and public role in edit form.
Adjusted authorization migration script to retain publish and update permissions related to public role.
Created migration for instances which have already updated to access control configuration version 1, which is part of VIVO 1.15.0

# How should this be tested?
Try migration for VIVO instances earlier than 1.15.0 release. Verify that publish permissions related to public role are preserved and could be modified.
For VIVO 1.15.0 this fix should allow to assign publish property permissions to public role. 

# Interested parties
@VIVO-project/vivo-committers

# Reviewers' expertise

1. Java
2. N3
3. Vitro access ontology

# Reviewers' report template
**Please update the following template which should be used by reviewers.**
## General comment
## Testing
## Code reviewing

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/migration/auth/AuthMigrator.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/migration/auth/AuthMigrator.java
@@ -73,7 +73,30 @@ public class AuthMigrator implements ServletContextListener {
         if (!isMigrationRequired()) {
             return;
         }
-        runCompleteMigration(sce, begin);
+        long currentVersion = getVersion();
+        if (currentVersion == 0) {
+            runCompleteMigration(sce, begin);
+        } else if (currentVersion == 1) {
+            migratePublishPublicPermissions(sce, begin);
+        }
+    }
+
+    private void migratePublishPublicPermissions(ServletContextEvent sce, long begin) {
+        ServletContext ctx = sce.getServletContext();
+        StartupStatus ss = StartupStatus.getBean(ctx);
+        log.info("Started publish permissions authorization reconfiguration for public role");
+        convertPublicPublishPermissions();
+        ss.info(this, secondsSince(begin) + " seconds spent to reconfigure publish permissions for public role");
+        removeVersion(getVersion());
+        setVersion(CURRENT_VERSION);
+        log.info(String.format("Updated access control configuration to version %d", CURRENT_VERSION));
+        PolicyLoader.getInstance().loadPolicies();
+        log.info("Reloaded all policies after migration");
+    }
+
+    private void convertPublicPublishPermissions() {
+        AnnotationMigrator annotationMigrator = new AnnotationMigrator(contentRdfService, configurationRdfService);
+        annotationMigrator.updatePublicPublishPermissions();
     }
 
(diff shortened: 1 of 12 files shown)


brianjlowe:
The migration from 1.14 doesn't seem to be working for me.  Here are the steps I did:
1. Clean install of 1.14.3 with empty tdbContentModels
2. Edited object property "access provided by (scires)"
3. Set publish permission to "curator and above"; save property.
4. Create new object property.
5. Set publish permission to "editor and above"; save property.
6. Clean install of PR.
7. Open "access provided by".
8. Publish permission is set to "public" (**should be curator and above**)
9. Open new property I created.
10. No permissions checkboxes are checked (**should be publish to editor and above; public display/edit**)

litvinovg:
First step should be to clean tdbConfigurationModels, not tdbContentModels

brianjlowe:
Repeating the above steps with both empty tdbModels and tdbContentModels at start improves one aspect.  Now the new property I added in step 4 has permissions checked, but the publish permissions for both this and the "access provided by" properties are still wrong.  After the upgrade using the PR, I have all the publish checkboxes checked for both properties, when one should be only curator and above and the other should be only editor and above.

brianjlowe (dismissed):
Tested upgrade from 1.14.3 and editing of publish flags in 1.15.  The object and data properties now appear in the public linked data where enabled.  The only thing I noticed in testing is that we lack rdf:type and rdfs:label in the public linked data, but they appear when the linked data is requested by a root user.  This makes some sense, since there is no place to edit permissions for these properties in the backend editor, but these are probably the two most useful things to include in linked data.  Perhaps they were always published before regardless of role level?  If that's the case and it's easy to replicate that behavior here, maybe it would make sense to include in this PR.  Otherwise, we can open a separate issue.

litvinovg:
Rebased PR on top of main branch. Added publish permissions for rdfs:label and rdf:type properties.
Created linked VIVO PR https://github.com/vivo-project/VIVO/pull/4054

api/src/main/java/edu/cornell/mannlib/vitro/webapp/migration/auth/AuthMigrator.java
```diff
@@ -72,6 +73,33 @@ public void contextInitialized(ServletContextEvent sce) {
         if (!isMigrationRequired()) {
             return;
         }
+        long currentVersion = getVersion();
```
chenejac, line 76: Can you please refactor your code that getVersion() is invoked only once? You are executing the same Sparql query in some cases three times.

chenejac (changes requested):
@litvinovg please check my comment.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/migration/auth/AuthMigrator.java
litvinovg: fixed

chenejac (approved):

PR #485, opened 2025-01-16
Created custom css style upload

milospp:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/4019)**
**[Linked Vivo PR](https://github.com/vivo-project/VIVO/pull/4039)**

This pull request enables administrators to manage custom CSS styles directly from the Site Admin page. Admins can upload, download, and remove custom CSS styles, as well as reset styles to ensure accessibility and functionality remain intact.

# What's new?

1. Added a new page for managing custom CSS styles under Site Admin > Site Information.
2. Features include:
    - Upload custom CSS files to override existing styles.
    - Download current custom CSS.
    - Remove custom CSS to revert to the default styles.
3. Added download links for CSS files used in the current theme.
4. Created a Reset button that remains visible and functional, even when styles are broken due to invalid or problematic CSS.

## Additional Notes
This PR introduces ReferrerHelper.java
This util class easily retrieves the previous page from the request and uses it for the back button.
The traditional approach for handling the back button was to always navigate to the document.referrer or previously visited page. However, this strategy breaks when dealing with nested navigation flows.
`Home -> Page 2 -> Page 3`
when the user is on Page 3 and clicks the back button, they are correctly taken back to Page 2. But at that point, the referrer becomes Page 3 again. So if the user clicks back on Page 2, they are taken to Page 3—creating an infinite loop between Page 2 and Page 3.

This ReferrerHelper class mitigates the problem by preventing the referrer from being overwritten in such nested navigation flows. 

# Screenshoots

### Style edit page
image
image

### Example css
```
body {
    background: #333;
}
```
image

### Broken contrast css (RESET button is still visible)
image

# How should this be tested?
_Test for every theme_
## Uploading regular CSS
1. Login as an admin.
2. Navigate to **Site Admin** > **Site Information** > **Change CSS Styles**.
3. Test downloading the default styles.
4. Upload a custom CSS file with the following content
```
body {
    background: red;
}
```

5. Clikc upload
_Expected Result: The site background should change to red._

## Download Custom CSS
1. After uploading custom CSS, return to the **Change CSS Styles** page.
2. Click **Download Custom CSS**.
_Expected Result: The downloaded file should match the uploaded custom CSS._

## Removing Custom CSS

After uploading custom CSS, navigate to the **Change CSS Styles** page.
Click **Remove Custom CSS**.
_Expected Result: The site should revert to the default styles._

## Uploading broken CSS
1. Test with CSS that introduces accessibility or visibility issues, such as
Example:
```
* {
    background: white !important;
    color: white !important;
    border-color: transparent !important;
}
```
or
```
* {
    background: white !important;
    color: white !important;
    border-color: transparent !important;
    display: none !important;
    opacity: 0 !important;
    visibility: hidden !important;
}
```
2. Upload the CSS file and observe the Reset button.
_Reset button should still be visible and working_
3. Click Reset

# Additional Notes:
The custom CSS file is saved using the Vitro **FileStorage** class and file is stored into **vivo-home/uploads**.
The file's path is stored in the **ApplicationBean** under the key **customCssPath** (_vitro-kb-applicationMetadata_).

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/SiteStyleController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/SiteStyleController.java
@@ -4,6 +4,9 @@ package edu.cornell.mannlib.vitro.webapp.controller.freemarker;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 
 import javax.servlet.ServletException;
 import javax.servlet.annotation.WebServlet;
@@ -66,6 +69,7 @@ public class SiteStyleController extends FreemarkerHttpServlet {
     public static boolean customCssUrlLoaded = false;
     public static String customCssUrl = null;
     public static String customCssFileUri = null;
+    public static String customCssVersion = null;
 
 
     private FileStorage fileStorage;
@@ -247,6 +251,8 @@ public class SiteStyleController extends FreemarkerHttpServlet {
 
         String propertyUriFileUri = VitroVocabulary.PORTAL_CUSTOMCSSFILEURI;
         Property propertyFileUri = ResourceFactory.createProperty(propertyUriFileUri);
+        String propertyUriVersion = VitroVocabulary.PORTAL_CUSTOMCSSVERSION;
+        Property propertyVersion = ResourceFactory.createProperty(propertyUriVersion);
 
         if (!cssFilePath.isEmpty()) {
             Statement statementImageUrl =
@@ -256,8 +262,15 @@ public class SiteStyleController extends FreemarkerHttpServlet {
                 ResourceFactory.createTypedLiteral(fileInfo.getUri()));
             displayModel.add(statementUri);
 
+            String versionValue =
+                DateTimeFormatter.ofPattern("yyyyMMddHHmmss").format(ZonedDateTime.now(ZoneOffset.UTC));
+            Statement statementVersion = new StatementImpl(portalResource, propertyVersion,
+                ResourceFactory.createTypedLiteral(versionValue));
+            displayModel.add(statementVersion);
+
             SiteStyleController.customCssUrlLoaded = true;
-            SiteStyleController.customCssUrl = cssFilePath;
+            SiteStyleController.customCssVersion = versionValue;
+            SiteStyleController.customCssUrl = appendVersion(cssFilePath, versionValue);
             SiteStyleController.customCssFileUri = fileInfo.getUri();
 
         }
@@ -272,6 +285,10 @@ public class SiteStyleController extends FreemarkerHttpServlet {
 
         Property property = ResourceFactory.createProperty(VitroVocabulary.PORTAL_CUSTOMCSSPATH);
         displayModel.removeAll(themeResource, property, null);
+        Property propertyFileUri = ResourceFactory.createProperty(VitroVocabulary.PORTAL_CUSTOMCSSFILEURI);
+        displayModel.removeAll(themeResource, propertyFileUri, null);
+        Property propertyVersion = ResourceFactory.createProperty(VitroVocabulary.PORTAL_CUSTOMCSSVERSION);
+        displayModel.removeAll(themeResource, propertyVersion, null);
 
         String url = SiteStyleController.getCustomCssUrlCache();
         String fileUri = SiteStyleController.customCssFileUri;
@@ -291,11 +308,26 @@ public class SiteStyleController extends FreemarkerHttpServlet {
         Property customCssPathProperty = ResourceFactory.createProperty(VitroVocabulary.PORTAL_CUSTOMCSSPATH);
         Property customCssFileUriProperty = ResourceFactory.createProperty(
             VitroVocabulary.PORTAL_CUSTOMCSSFILEURI);
+        Property customCssVersionProperty = ResourceFactory.createProperty(
+            VitroVocabulary.PORTAL_CUSTOMCSSVERSION);
 
-        customCssUrl = getLiteralProperty(displayModel, styleResource, customCssPathProperty);
+        String basePath = getLiteralProperty(displayModel, styleResource, customCssPathProperty);
         customCssFileUri = getLiteralProperty(displayModel, styleResource, customCssFileUriProperty);
+        customCssVersion = getLiteralProperty(displayModel, styleResource, customCssVersionProperty);
+        customCssUrl = appendVersion(basePath, customCssVersion);
 
         customCssUrlLoaded = true;
+    }
+
+    private static String appendVersion(String basePath, String version) {
+        if (basePath == null || basePath.isEmpty()) {
+            return basePath;
+        }
+        if (version == null || version.isEmpty()) {
+            return basePath;
+        }
+        String separator = basePath.contains("?") ? "&" : "?";
+        return basePath + separator + "version=" + version;
     }
 
     private static String getLiteralProperty(OntModel model, Resource subject, Property property) {
@@ -316,6 +348,7 @@ public class SiteStyleController extends FreemarkerHttpServlet {
         customCssUrlLoaded = false;
         customCssUrl = null;
         customCssFileUri = null;
+        customCssVersion = null;
     }
 
     public static String getCustomCssUrlCache() {
--- a/webapp/src/main/webapp/templates/edit/specific/applicationBean_retry.jsp
+++ b/webapp/src/main/webapp/templates/edit/specific/applicationBean_retry.jsp
@@ -43,7 +43,7 @@
         <br><br>
         
         <label for="fileUpload" class="">${i18n.text('custom_css_style')}</label>
-        <input type="file" id="fileUpload" name="fileUpload" accept=".css" style="margin-bottom: 10px; display: ${not empty customCssPath and customCssPath != 'null' ? 'none' : 'inline'};">
+        <input type="file" id="fileUpload" name="fileUpload" accept=".css" class="css-file-upload ${not empty customCssPath and customCssPath != 'null' ? 'hidden' : ''}">
         <c:if test="${not empty customCssPath and customCssPath != 'null'}">
             <br>
             <button id="resetAction" type="button" class="button red" onclick="resetStyles()">${i18n.text('remove_custom_css')}</button>
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/VitroVocabulary.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/dao/VitroVocabulary.java
@@ -124,6 +124,7 @@ public class VitroVocabulary {
     public static final String PORTAL_COPYRIGHTANCHOR = vitroURI+"copyrightAnchor";
     public static final String PORTAL_CUSTOMCSSPATH = vitroURI+"customCssPath";
     public static final String PORTAL_CUSTOMCSSFILEURI = vitroURI+"customCssFileUri";
+    public static final String PORTAL_CUSTOMCSSVERSION = vitroURI+"customCssVersion";
 
 
     // reusing displayRank property above
--- a/webapp/src/main/webapp/css/vitro.css
+++ b/webapp/src/main/webapp/css/vitro.css
@@ -532,6 +532,10 @@ a.button.blue {
     background-color: #b40000
 }
 
+.css-file-upload {
+    margin-bottom: 10px;
+}
+
 .clear-file-btn {
     margin-left: 10px;
     padding: 2px 6px;


balmas (changes requested):
On the whole, this works as documented.  

The CSS links on the Style upload page are broken when VIVO is deployed at the root and not /vivo.

Also there is some security risk with this feature as a whole, since it's possible that CSS injection could be used in the uploaded css file (or potentially other malware).  I suppose this is mitigated by the fact that only an admin has access to the page, but it still feels worth pointing out.   

I wonder if it would be better to allow the user to specify a remote url for css rather than uploading the file locally.

webapp/src/main/webapp/templates/freemarker/body/siteAdmin/siteAdmin-siteStyle.ftl
```diff
@@ -0,0 +1,130 @@
+<div>
+    <h2>Site Styles</h2>
+    <div class="button-group">
```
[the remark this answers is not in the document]
milospp: Great catch, I'll fix this right away

balmas (approved):
With the fix to the css links in place, I'm good with this PR now.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/FreemarkerHttpServlet.java
```diff
@@ -469,6 +476,9 @@ protected Map<String, Object> getPageTemplateValues(VitroRequest vreq) {
 
         // Add these accumulator objects. They will collect tags so the template can write them
         // at the appropriate location.
+
+        map.putAll(getCustomCss());
```
litvinovg, line 480: I think instead you could create a static String field initialized with the same code and updated by SiteStyleController.

webapp/src/main/webapp/templates/freemarker/body/siteAdmin/siteAdmin-siteStyle.ftl
```diff
@@ -0,0 +1,130 @@
...
+        <a class="button blue" download href="${urls.base}/css/login.css">login.css</a>
+        <a class="button blue" download href="${urls.base}/css/edit.css">edit.css</a>
+        <a class="button blue" download href="${urls.base}/css/individual/individual-property-groups.css">individual-property-groups.css</a>
+        <a class="button blue" download href="${urls.base}/css/developer/developerPanel.css">developerPanel.css</a>
+        
+        <!-- Samo Wilma i TF-->
+        <a class="button blue" download href="${urls.base}/css/home-page-maps.css">page-createAndLink.css</a> 
+
+
+        <#if themeDir == "themes/wilma">
```
litvinovg, line 14: Hardcoding theme paths is not an option. It will be broken once theme is renamed for customization.

webapp/src/main/webapp/templates/edit/specific/applicationBean_retry.jsp
```diff
@@ -40,6 +40,10 @@
             <form:option name="ThemeDir" />
         </select>
 
+        <br><br>
+        <a href="${siteStyleUrl}">Change CSS Styles</a>
```
balmas, line 44: I suggest rewording this text to "Add CSS Style Overrides"

webapp/src/main/webapp/templates/freemarker/body/siteAdmin/siteAdmin-siteStyle.ftl
```diff
@@ -0,0 +1,72 @@
...
+   <#if customCssPath?? && customCssPath != "null">
+       <button
+       style="font-weight: normal !important; color: white !important; background-color: #b40000 !important; font-size: 16px !important; display: inline !important; visibility: visible !important; opacity: 1 !important; z-index: 10 !important; width: 201px !important; height: 31px !important;"
+        id="resetAction" type="button" class="button red" onclick="resetStyles()">Remove Custom CSS</button>
+       <a href="${customCssPath}" download="custom.css" class="button blue">Download Custom CSS</a>
+   </#if>
+   <form action="${actionUpload}" method="post" enctype="multipart/form-data">
+       <label for="fileUpload" class="">Upload new custom style.css:</label>
+       <input type="file" id="fileUpload" name="fileUpload" accept=".css" class="form-item">
+       <p class="note">Note: Avoid using relative links inside the CSS file to ensure proper loading of resources.</p>
```
balmas, line 16: this note is a little strange.  I would think relative links would work better than absolute ones. What is the reason for it?

webapp/src/main/webapp/templates/freemarker/body/siteAdmin/siteAdmin-siteStyle.ftl
```diff
@@ -0,0 +1,72 @@
...
+   <h3>Custom CSS Style</h3>
+
+   <#if customCssPath?? && customCssPath != "null">
+       <button
+       style="font-weight: normal !important; color: white !important; background-color: #b40000 !important; font-size: 16px !important; display: inline !important; visibility: visible !important; opacity: 1 !important; z-index: 10 !important; width: 201px !important; height: 31px !important;"
+        id="resetAction" type="button" class="button red" onclick="resetStyles()">Remove Custom CSS</button>
+       <a href="${customCssPath}" download="custom.css" class="button blue">Download Custom CSS</a>
+   </#if>
+   <form action="${actionUpload}" method="post" enctype="multipart/form-data">
+       <label for="fileUpload" class="">Upload new custom style.css:</label>
```
balmas, line 14: the file doesn't need to be named style.css so maybe change this text to "Upload new css file" ?

balmas (approved):
This seems to work well enough. My review is mostly of the functionality and not of the code itself. It would be good to have someone with more familiarity with the VIVO code to confirm the code.  I did have a couple of minor suggestions for wording changes for clarity but they aren't critical.

webapp/src/main/webapp/templates/edit/specific/applicationBean_retry.jsp
milospp: Updated

webapp/src/main/webapp/templates/freemarker/body/siteAdmin/siteAdmin-siteStyle.ftl
milospp: In the previous version, the feature included links to download all CSS files with styles that used relative paths for assets (e.g., url(../logo.png)). This note is just a reminder: if you're copying styles from those files, make sure to update any relative paths, since custom CSS is now stored in a different location and those paths won't work as-is.

However, since the list of downloadable CSS files was removed in the last commit, this note is now confusing from a user’s perspective — so I’ll delete it.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/FreemarkerHttpServlet.java
```diff
@@ -20,6 +20,9 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.jena.ontology.OntModel;
+import org.apache.jena.rdf.model.*;
```
litvinovg, line 24: Could you fix wildcard import here?

home/src/main/resources/rdf/i18n/en_US/interface-i18n/firsttime/vitro_UiLabel.ttl
```diff
@@ -6931,3 +6931,34 @@ uil-data:there_are_no_results_to_display.Vitro
         uil:hasKey       "there_are_no_results_to_display" ;
         uil:hasPackage   "Vitro-languages" .
 
+uil-data:back.Vitro
```
litvinovg, line 6934: I think it is not used.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/SiteStyleController.java
```diff
@@ -0,0 +1,295 @@
...
+
+        if (!cssFilePath.isEmpty() && !cssFilePath.equals("null")) {
+            Statement statement =
+                new StatementImpl(portalResource, property, ResourceFactory.createTypedLiteral(cssFilePath));
+            displayModel.add(statement);
+        }
+    }
+
+
+    private void removeCssFileDisplayModel() {
```
litvinovg, line 250: Uploaded css style file is not deleted on removal.

litvinovg: File related triples should be removed too.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/SiteStyleController.java
```diff
@@ -0,0 +1,295 @@
...
+        OntModel displayModel = cma.getOntModel(ModelNames.DISPLAY);
+
+        Resource portalResource = ResourceFactory.createResource(VitroVocabulary.PROPERTY_CUSTOMSTYLE);
+
+        String propertyUri = VitroVocabulary.PORTAL_CUSTOMCSSPATH;
+        Property property = ResourceFactory.createProperty(propertyUri);
+
+        displayModel.removeAll(portalResource, property, null);
+
+        if (!cssFilePath.isEmpty() && !cssFilePath.equals("null")) {
```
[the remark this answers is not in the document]
milospp: It looks like this isn't used, I initially added it to improve the robustness of the function, but on closer inspection, it seems that workflow is never going to happen.
Removed that check

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/SiteStyleController.java
```diff
@@ -0,0 +1,295 @@
...
+
+    private void removeCssFile(VitroRequest vreq) {
+        removeCssFileDisplayModel();
+        resetCustomCssCache();
+    }
+
+    private void showMainStyleEditPage(HttpServletRequest request, HttpServletResponse response, String info) {
+        try {
+            response.setContentType("text/html");
+            response.getWriter().println("<html><body><h2>Site Style</h2><pre>customCssUrl: "
```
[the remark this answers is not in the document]
litvinovg: How this html is used in direct API calls? I couldn't find the code.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/SiteStyleController.java
milospp: Added deleting file on replace or remove CSS

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/SiteStyleController.java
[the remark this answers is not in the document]
milospp: To check what file is uploaded
http://localhost:8080/vivo/siteStyle
image

To remove file (works only if admin is logged in)
http://localhost:8080/vivo/siteStyle?action=remove
image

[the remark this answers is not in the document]
milospp: This is the same request that is sent when the user clicks on the remove button, but in case somehow this is not available, this is a workaround

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/SiteStyleController.java
```diff
@@ -0,0 +1,317 @@
...
+        return vreq.getFiles().get(PARAMETER_UPLOADED_FILE).get(0);
+    }
+
+    private FileInfo createFile(FileItem file, String storedFileName, UploadedFileHelper fileHelper)
+        throws FileUploadController.FileUploadException {
+        FileInfo fileInfo;
+        try {
+            fileInfo = fileHelper.createFile(storedFileName, getMediaType(file), file.getInputStream());
+        } catch (Exception e) {
+            log.error(e.getLocalizedMessage());
```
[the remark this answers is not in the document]
milospp: Repalced

webapp/src/main/webapp/templates/edit/formBasic.js
```diff
@@ -1,7 +1,74 @@
...
-function confirmDelete() {
-    var msg="Are you SURE you want to delete this record? If in doubt, CANCEL."
-    return confirm(msg);
-}
+    function confirmDelete() {
+        var msg="Are you SURE you want to delete this record? If in doubt, CANCEL."
+        return confirm(msg);
+    }
+
+    async function onSave(e) {
```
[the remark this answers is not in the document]
litvinovg: Could new js be added in applicationBean_retry.jsp ?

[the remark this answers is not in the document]
milospp: I can add a <script> tag inside `applicationBean_retry.jsp`, which would inject the script into the body. This approach works technically, and it aligns with how scripts are already being included in some other parts of VIVO. However, my only concern is whether it's clean practice to place script tags in the body instead of the head, as is typically recommended.

Currently in `formBasic.js`, there is only this code, which I think is not used in this jsp page, therefore I can also replace the script inside Controller without problems.
```
<!-- $This file is distributed under the terms of the license in LICENSE$ -->
<script language="JavaScript" type="text/javascript">
function confirmDelete() {
    var msg="Are you SURE you want to delete this record? If in doubt, CANCEL."
    return confirm(msg);
}
</script>
```

What do you think?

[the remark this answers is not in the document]
litvinovg: if it is not used on the page, then sure it should be easier to just replace formBasic.js with new js filename.

litvinovg:
@milospp SiteStyleController provides unrestricted access to upload/remove css styles.
Code to check authorization is being overridden by current implementation.

balmas:
Some things in the functionality have changed in this latest version:

1) When uploading a file with css that is invalid (e.g. as in the example in the testing instructions which sets the visibility of everything to hidden) the Reset and Delete buttons no longer appear.  You can still use the browser's back button to get out of this, so I'm not sure it's a critical issue, and it seems like an edge case to me, so I'm not sure that this is a critical show stopper.  However, I'm pointing it out since it's different than the described functionality. If we no longer think that functionality is important we should update the description of the PR.

2) There is a difference in functionality in Firefox and Chrome.  In Firefox, when you click Save after uploading a CSS file, you get an error in the browser console: "Form contains a file input, but is missing method=POST and enctype=multipart/form-data on the form.  The file will not be sent. editForm:52:67". And then you are redirected to the main VIVO page.  It does appear that the css is uploaded, because the CSS change takes hold but the error is a little worrisome.   In Chrome, you don't get that error, and upon save instead of returning to the main VIVO screen, you stay on the Site Information page after upload.  Safari works like Chrome.

3) One thing that's a little confusing is the "No file selected" (or "No file chosen" depending on the browser) that appears on this form next to the File Upload button, even once you have uploaded a CSS file.  I realize that this may be a browser default for the behavior of that button, but it does feel a little misleading to see that after a successful file upload.  Maybe it would be helpful to just display a message that says something like "<filename> successfully uploaded" to reassure the user that their file is in place.

home/src/main/resources/rdf/i18n/en_US/interface-i18n/firsttime/vitro_UiLabel.ttl
```diff
@@ -6955,6 +6955,14 @@ uil-data:upload_new_css.Vitro
         uil:hasKey       "upload_new_css" ;
         uil:hasPackage   "Vitro-languages" .
 
+uil-data:cusotm_css_style.Vitro
+        rdf:type         owl:NamedIndividual ;
+        rdf:type         uil:UILabel ;
+        rdfs:label       "Custom css file"@en-US ;
+        uil:hasApp       "Vitro" ;
+        uil:hasKey       "cusotm_css_style" ;
```
balmas, line 6963: @milospp could you just correct the typo on the key name ("custom_css_style" instead of cusotm_css_style") ?

webapp/src/main/webapp/js/siteAdmin/cssUploadUtils.js
```diff
@@ -46,29 +46,119 @@
...
+    function createClearButton(fileInput) {
+        const clearBtn = document.createElement('button');
+        clearBtn.id = 'clearFileBtn';
+        clearBtn.type = 'button';
+        clearBtn.innerHTML = '✕';
+        clearBtn.title = 'Remove selected file';
+        
+        // Style the clear button
+        Object.assign(clearBtn.style, {
+            marginLeft: '10px',
```
litvinovg, line 89: Wouldn't it be better to move css styles into css file and assign them by class name?

PR #491, opened 2025-02-13
Avoid invalid substitutions in translation converter

litvinovg:
Avoid possible substitutions in translation body if body content matches sparql variable name.
For example if translation string contains a variable with a link 
 <a href=\"/fis/individual?uri=http://d-nb.info/gnd/1080328793\">
 ?uri part match variable name and is being substituted with variable value.
 To prevent that new parameterized sparql string is being created before substitution of label value.

# How should this be tested?

- Run extended translation converter test

- Manual testing:
1. Create a translation in properties file with body that contains ?uri text
2. Start VIVO to run translation conversions
3. Check created translation in http://vitro.mannlib.cornell.edu/default/interface-i18n  graph

# Interested parties
@VIVO-project/vivo-committers

# Reviewers' expertise
1. Java
# Reviewers' report template
## General comment
## Testing
## Code reviewing


chenejac (approved):
Before application of the fix (in TranslationConverter.java) the tests are failing, after application of the fix, the tests are passing. @litvinovg thanks for this. As we discussed at the last committers meeting one review/test of this PR might be enough for merging taking into account its size.

PR #504, opened 2025-12-17
Feature/integration model

ivanmrsulja:
Issues: [4074](https://github.com/vivo-project/VIVO/issues/4074), [4075](https://github.com/vivo-project/VIVO/issues/4075)

linked VIVO PR: [4125](https://github.com/vivo-project/VIVO/pull/4125)

Added support for integration model graph, where integration metadata (access tokens, push history, utility metadata) can be stored for various external platforms.

# What's new?
- api/src/main/java/edu/cornell/mannlib/vitro/webapp/servlet/setup/ConfigurationModelsSetup.java
- api/src/main/java/edu/cornell/mannlib/vitro/webapp/modules/tripleSource/ConfigurationTripleSource.java
- api/src/main/java/edu/cornell/mannlib/vitro/webapp/modelaccess/ModelNames.java

# How should this be tested?
This should be tested in conjunction with [corresponding VIVO PR](https://github.com/vivo-project/VIVO/pull/4125) that introduces Orcid Member API integration for push workflows. It uses this model for integration metadata storage, if it works as expected, so does this.

# Interested parties
@chenejac 

# Reviewers' expertise

1. Java


webapp/src/main/webapp/css/vitro.css
```diff
@@ -543,3 +543,12 @@ a.button.blue {
...
 }
+
+.mt-5 {
+    margin-top: 3.5rem;
+}
+
+.section-title-small {
+    margin-bottom: 0rem !important;
+}
+
```
chenejac: What is the purpose of those lines? Why !important? Can that be changed in themes if it is defined as here? Morover, ORCID integration is VIVO specific task, not sure why css configuration is in Vitro code.

chenejac (changes requested):

webapp/src/main/webapp/css/vitro.css
ivanmrsulja: I consulted @milospp and he suggested to move this CSS to VIVO's `wilma.css`, so I did that. Also, `!important` is not needed anymore because, after moving the code, load order is as it should be.

PR #511, opened 2026-03-04
Fixed critical ES connection pooling bug

ivanmrsulja:
[VIVO Issue](https://github.com/vivo-project/VIVO/issues/4144)
Fixes ES connection pooling bug where used connections are not closed, resulting in connection exhaustion.

# What's new?
Added relevant connection closing logic.

# How should this be tested?
Try to index/reindex collection with more than 500 records. It should not freeze up and it should complete successfully.

# Interested parties
@chenejac 

# Reviewers' expertise

1. Java

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ESAdder.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ESAdder.java
@@ -3,6 +3,7 @@
 package edu.cornell.mannlib.vitro.webapp.searchengine.elasticsearch;
 
 import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -19,10 +20,10 @@ import edu.cornell.mannlib.vitro.webapp.search.VitroSearchTermNames;
 import edu.cornell.mannlib.vitro.webapp.utils.http.ESHttpBasicClientFactory;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.apache.http.HttpResponse;
-import org.apache.http.client.HttpClient;
+import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPut;
 import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.util.EntityUtils;
 
 /**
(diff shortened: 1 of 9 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ESAdder.java
```diff
@@ -114,19 +115,20 @@ private void putToElastic(String json, String docId)
             throws SearchEngineException {
         try {
             String url = baseUrl + "/_doc/"
-                    + URLEncoder.encode(docId, "UTF8");
-            HttpClient httpClient = ESHttpBasicClientFactory.getHttpClient(baseUrl);
+                    + URLEncoder.encode(docId, StandardCharsets.UTF_8);
+            CloseableHttpClient httpClient = ESHttpBasicClientFactory.getHttpClient(baseUrl);
```
litvinovg, line 119: Shouldn't CloseableHttpClient instance be closed at some point?

litvinovg: The same issue appears in all other changes. Please fix it.

ivanmrsulja: `HttpClient` instances are shared and backed by a pooling connection manager (`setConnectionManagerShared(true)`), so they are intentionally not closed per request.
That said, I should probably create a shutdown hook to properly close it on application termination.

ivanmrsulja: Done 😃 !

litvinovg: It is not safe to use shared closable instance of HttpClient.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/searchengine/elasticsearch/ESFlusher.java
```diff
@@ -31,10 +31,11 @@ public void flush(boolean wait) throws SearchEngineException {
         try {
             String url = baseUrl + "/_flush"
                     + (wait ? "?wait_for_ongoing" : "");
-            HttpClient httpClient = ESHttpBasicClientFactory.getHttpClient(baseUrl);
-            HttpResponse response = httpClient.execute(new HttpGet(url));
-            String json = EntityUtils.toString(response.getEntity());
-            log.debug("flush response: " + json);
+            CloseableHttpClient httpClient = ESHttpBasicClientFactory.getHttpClient(baseUrl);
+            try (CloseableHttpResponse response = httpClient.execute(new HttpGet(url))) {
```
litvinovg, line 35: Exceptions in this try will not be caught.

litvinovg:
Taken into account that we need to have the fix as soon as possible for the release maybe it would be faster if you could find time to check this solution https://github.com/vivo-project/Vitro/compare/main...litvinovg:Vitro:vivo-es-http-client-fix
It has single EsHttpClient which sends requests to ElasticSearch with execute method without exposing ClosableHttpClient instance to any external code, which might accidentally close it. Also removed duplicated http client instance, removed try in which exceptions will not be caught and simplified creation of http client instance.
If it looks good to you @ivanmrsulja  and @chenejac I can create a PR from that commit.

ivanmrsulja:
> Taken into account that we need to have the fix as soon as possible for the release maybe it would be faster if you could find time to check this solution [main...litvinovg:Vitro:vivo-es-http-client-fix](https://github.com/vivo-project/Vitro/compare/main...litvinovg:Vitro:vivo-es-http-client-fix) It has single EsHttpClient which sends requests to ElasticSearch with execute method without exposing ClosableHttpClient instance to any external code, which might accidentally close it. Also removed duplicated http client instance, removed try in which exceptions will not be caught and simplified creation of http client instance. If it looks good to you @ivanmrsulja and @chenejac I can create a PR from that commit.

I took a look at your solution and I agree it’s simpler. Having a single `EsHttpClient` with an execute method and not exposing the `CloseableHttpClient` does reduce the risk of accidental misuse (like closing it externally), and the overall flow is easier to follow. That said, there is a tradeoff to be aware of in my opinion, that this approach introduces some rigidity (harder to reconfigure or extend later). Also, both implementations currently trust all SSL certificates, which is something we should definitely revisit in the future.

In the end I’m fine with going with your approach :smile:

chenejac:
superseded by #512

PR #513, opened 2026-04-07
WCAG - Miscelanious

milospp:
**[VIVO GitHub issue](https://github.com/vivo-project/VIVO/issues/4103)**
**[VIVO PR](https://github.com/vivo-project/VIVO/pull/4138)**

# What's new?
Improved HTML title, title is dynamic based on the selected page in the RDF
Added a missing title in some input fields in the profile
image

# Reviewers' expertise

1. Java
1. HTML, CSS, JavaScript
1. Natural language knowledge

### DIFF
--- a/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/HomePageController.java
+++ b/api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/HomePageController.java
@@ -53,7 +53,8 @@ public class HomePageController extends FreemarkerHttpServlet {
         }*/
         body.put("dataServiceUrlVClassesForVClassGroup", UrlBuilder.getUrl("/dataservice?getVClassesForVClassGroup=1&classgroupUri="));
 	    body.put("geoFocusMapsEnabled", getGeoFocusMapsFlag(vreq));
-
+        body.put("title","Home");
+        
         return new TemplateResponseValues(BODY_TEMPLATE, body);
     }
 
--- a/webapp/src/main/webapp/templates/freemarker/edit/forms/js/customFormWithAutocomplete.js
+++ b/webapp/src/main/webapp/templates/freemarker/edit/forms/js/customFormWithAutocomplete.js
@@ -557,6 +557,7 @@ var customForm = {
         // boolean controls whether the acSelector value gets cleared.
 
         var clearAcSelectorVal = true;
+        let focus = null;
 
         if ( $(selectedObj).attr('id') == "typeSelector" ) {
             $acSelectionObj = customForm.acSelections[$(selectedObj).attr('acGroupName')];
--- a/webapp/src/main/webapp/templates/freemarker/lib/lib-properties.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/lib/lib-properties.ftl
@@ -348,3 +348,7 @@ name will be used as the label. -->
     <#local groupName = groupName?replace("&", "-and-")>
     <#return groupName>
 </#function>
+
+<#function capitalizeGroupName propertyGroupName>
+    <#return propertyGroupName?cap_first>
+</#function>
--- a/webapp/src/main/webapp/templates/freemarker/page/partials/title.ftl
+++ b/webapp/src/main/webapp/templates/freemarker/page/partials/title.ftl
@@ -1,3 +1,7 @@
 <#-- $This file is distributed under the terms of the license in LICENSE$ -->
 
-<title>${(title?html)!siteName!}</title>
+<#if title?? && title?has_content>
+    <title>${title?html} | ${siteName}</title>
+<#else>
+    <title>${siteName}</title>
+</#if>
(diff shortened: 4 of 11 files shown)


api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/HomePageController.java
```diff
@@ -53,7 +53,8 @@ protected ResponseValues processRequest(VitroRequest vreq) throws InstantiationE
         }*/
         body.put("dataServiceUrlVClassesForVClassGroup", UrlBuilder.getUrl("/dataservice?getVClassesForVClassGroup=1&classgroupUri="));
 	    body.put("geoFocusMapsEnabled", getGeoFocusMapsFlag(vreq));
-
+        body.put("title","Home");
```
rasteginas, line 56: This solution only provides an English title of the homepage for all language translations, though Vivo already has a translated homepage title in different languages ​​(it can be configured in the admin panel - page management), which might be worth using as homepage title, to match the translation to the chosen language version.

webapp/src/main/webapp/templates/freemarker/page/partials/title.ftl
```diff
@@ -1,3 +1,7 @@
 <#-- $This file is distributed under the terms of the license in LICENSE$ -->
 
-<title>${(title?html)!siteName!}</title>
+<#if title?? && title?has_content>
+    <title>${title?html} | ${siteName}</title>
```
rasteginas, line 4: I have some doubts  that adding a vertical bar with the same site name after each page title is the good solution. Are these changes necessary to comply with WCAG standards?

webapp/src/main/webapp/templates/freemarker/edit/forms/js/customFormWithAutocomplete.js
```diff
@@ -557,6 +557,7 @@ var customForm = {
         // boolean controls whether the acSelector value gets cleared.
 
         var clearAcSelectorVal = true;
+        let focus = null;
```
rasteginas, line 560: Could you please explain what “let focus = null” is used for ?

webapp/src/main/webapp/templates/freemarker/lib/lib-properties.ftl
```diff
@@ -348,3 +348,7 @@ name will be used as the label. -->
     <#local groupName = groupName?replace("&", "-and-")>
     <#return groupName>
 </#function>
+
+<#function capitalizeGroupName propertyGroupName>
```
rasteginas, line 352: Why do we need function instead of using ?cap_first inline?

webapp/src/main/webapp/templates/freemarker/page/partials/title.ftl
milospp: https://www.w3.org/WAI/WCAG22/Understanding/page-titled.html

In this example, the W3C recommends using a vertical separator.

Other popular websites also use either a pipe (|) or a dash (-) to separate parts of the page title.
image

webapp/src/main/webapp/templates/freemarker/edit/forms/js/customFormWithAutocomplete.js
milospp: Somehow, the rest of the code disappeared while I was splitting the changes into multiple PRs. I’ll add the missing code.

This was intended for the following use case:
image

When the user clicks Change selection, the code clears the Organisation Type dropdown. For some reason, this also causes the dropdown to lose focus and the focus jumps back to the top.

This fix keeps the focus on the dropdown after its value is cleared.

api/src/main/java/edu/cornell/mannlib/vitro/webapp/controller/freemarker/HomePageController.java
milospp: I replaced this with the translated version and also updated other hard-coded English titles in the codebase.

webapp/src/main/webapp/templates/freemarker/page/partials/title.ftl
rasteginas: Perhaps I didn't express myself clearly. What I meant was that the purpose of displaying the site name on every page isn't clear. The vertical bar itself is beside the point. WСAG recommends keeping page titles as specific as possible so that users can easily identifying the page from the title. Duplicating the site name in every title contradicts with that in my opinion.

webapp/src/main/webapp/templates/freemarker/lib/lib-properties.ftl
milospp: Actually, I’ll switch to the inline version, since this abstraction ends up being more verbose than simply calling `?cap_first`.

webapp/src/main/webapp/templates/freemarker/page/partials/title.ftl
milospp: I don't think appending the site name contradicts the guidance either. It is recommended to put the unique and most relevant information first, followed by the site or organization name when appropriate.

So the intention here is to keep the page-specific information first while also providing the site context, e.g. John Doe | VIVO. This can also make tabs easier to identify when pages from multiple sites are open.

I have checked some popular website and they all use this standard (except github 😄 )

Do you have some suggestion how should that look like?

rasteginas: I would keep the design currently implemented in VIVO. If the title is longer than two words—or if many tabs are open in the browser—the site name isn't visible anyway, so we have to rely on the favicon icon; meanwhile, when we are hovering the cursor over the tab title we see a duplication of the site name and its URL address. In my view, this looks unnecessary.

rasteginas: And I noticed that the site name in the tab title is not displayed on the "Capability map" page.

PR #518, opened 2026-04-24
Modify I18nSort document modifier to work around Jena bug

brianjlowe:
**[VIVO-4147](https://github.com/vivo-project/VIVO/issues/4147)**:

Changes documentModifierI18nSort.n3 so that the MIN() aggregration operates only on plain strings as a workaround for a Jena bug in literal comparisons that sometimes causes MIN() to return null.

# How should this be tested?
If you were able to reproduce the bug in your instance with the sample data on the VIVO-4147 issue, then, after applying the PR, re-load the same data file and the "Doe, John" should appear alphabetized correctly in the appropriate browse pages. Otherwise, just check the query for correctness and rebuild the search index confirming that counts do not change.

# Interested parties
@VIVO-project/vivo-committers

# Reviewers' expertise

1. SPARQL

# Reviewers' report template
**Please update the following template which should be used by reviewers.**
## General comment
## Testing
## Code reviewing

### DIFF
--- a/home/src/main/resources/rdf/display/everytime/documentModifierI18nSort.n3
+++ b/home/src/main/resources/rdf/display/everytime/documentModifierI18nSort.n3
@@ -9,8 +9,11 @@
     :hasTargetSuffix "_label_sort" ;
     :hasSelectQuery """
         PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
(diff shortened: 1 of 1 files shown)


chenejac (approved):
@brianjlowe looks ok to me as a temporary fix. It is nice also that you left the comment in the case Jena fix this issue how to return back to simple solution.

